PR #20, opened 2017-06-06
fixed \Model\Image class name conflicts on \Model\Project\Category and…

bqevin:
What does this PR do
- Fixes \Model\Image class name conflict with alias on Model classes for Project\Category and Project\Reviews 
- Fixes #19

PR #21, opened 2017-06-06
Catch error on reviews page empty content

bqevin:
What does this PR do
- Fixes error thrown when navigate to reviews page as admin before assigning project review


microstudi:
That's ok for your last commit but it has the same changes as your other PR #20. It should have only the last commit or remove the other PR

PR #24, opened 2017-06-10
Hiding code information on 404 error

bqevin:


microstudi:
Not sure about this.
Msg info is usefull in some cases (developing), in "real" mode messages are hidden anyway.
If you really want to do this in your particular case, a plugin would be a better option.
It's very easy to overwrite views in a plugin, just create the plugin, activate it and put an alternate version of the template following the same directory structure.
Example:
`Resources/templates/default/errors/not_found.php`
Can be overwritten by a plugin:
`extend/my-plugin/Resources/templates/default/errors/not_found.php`

bqevin:
That's brilliant! Thanks  for the insight!

PR #25, opened 2017-06-12
Caught access denied error to admin dashboard

Reaper45:
Catches the access denied error when a user is redirected to /admin after registration.

### DIFF
--- a/src/Goteo/Controller/AdminController.php
+++ b/src/Goteo/Controller/AdminController.php
@@ -157,18 +157,19 @@ class AdminController extends \Goteo\Core\Controller {
 		];
 
 		// If menu is not allowed, throw exception
-		if (empty($menu) || ($option && !array_key_exists($option, $menu))) {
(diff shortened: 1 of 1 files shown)


microstudi:
Can you explain whay you want to do that?
i think throwing an exception will indicate why you are not allowed in this administration zone (Access Denied)

Reaper45:
This is the error!
screenshot from 2017-06-14 17-48-03

microstudi:
An your user is admin/superadmin? if not this is correct (the debuging lines are not shown in a production environtment).
Can you show me the result of this SQL query?:
`SELECT * FROM user_role WHERE user_id='ivan';`
Change the 'ivan' for your user id

Reaper45:
The table `user_role` is empty.

Reaper45:
My `user_id` is `joram`, and the record is not in the `user_role` table.

PR #86, opened 2018-03-15
fix: session start

asterion:


public/index.php
```diff
@@ -19,6 +19,8 @@
 
 require_once __DIR__ . '/../src/autoload.php';
 
+\Goteo\Application\Session::factory(); // session start
```
microstudi, line 22: Would you care to explain why to put this here?
This should be done automatically in the event listener `SessionListener`

PR #96, opened 2018-05-22
Move rewards required value to settings

alexm:
Even though 3 seems a fair value as the required amount of rewards
for any given project, it is better to set that value through the
global settings than having it hardcoded.

The former default is preserved if the settings value is not found.

### DIFF
--- a/src/Goteo/Model/Project.php
+++ b/src/Goteo/Model/Project.php
@@ -2506,9 +2506,11 @@ namespace Goteo\Model {
             } else {
                 $res->rewards = 100;
             }
-            if($total < 3) {
+            // rewards required >= 1, default 3
+            $rewards_required = abs(intval(Config::get('rewards.required'))) ?: 3;
+            if($total < $rewards_required) {
                 $errors['rewards'][] = 'rewards_required';
-                $res->rewards *= $total / 3;
+                $res->rewards *= $total / $rewards_required;
             }
 
             $campaign = [ ];
--- a/config/demo-settings.yml
+++ b/config/demo-settings.yml
@@ -55,6 +55,10 @@ url:
     #
     data:
 
+# Default required rewards
+rewards:
+    required: 3
+
 # PLUGINS
 # Extend is the directory to personalize your copy of goteo.org
 # Routes, classes and templates can be overridden by copying the main structure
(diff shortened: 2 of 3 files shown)


src/Goteo/Model/Project.php
```diff
@@ -2506,9 +2506,10 @@ public function getValidation() {
             } else {
                 $res->rewards = 100;
             }
-            if($total < 3) {
+            $rewards_required = Config::get('rewards.required') ?: 3;
+            if($total < $rewards_required) {
                 $errors['rewards'][] = 'rewards_required';
-                $res->rewards *= $total / 3;
+                $res->rewards *= $total / $rewards_required;
```
aaguilera: Beware: this could fail when $rewards_required is 0.

microstudi: Absolutely right. @alexm could you put a conditional to bypass checks in that case?

alexm: Please, note that `"0" ?: 3` expression returns `3`.

On the other side, `" " ?: 3` returns space, so it would be possible to have a divide by zero error in that case. But, how probable is that anybody uses a space as a value for that setting?

Anyway, I guess I can improve it like this and it should work for all cases:

```
$rewards_required = intval(Config::get("rewards.required")) ?: 3;
```

alexm:
Improved the expression for $rewards_required. Thanks!

microstudi:
In the latests version of the code I decided to change the implementation of the validation code.
Now it's done via the EventDispatcher.
This even is thrown in the [Project model](https://github.com/GoteoFoundation/goteo/blob/devel/src/Goteo/Model/Project.php#L2280) and the validation class is [ProjectValidationEvent](https://github.com/GoteoFoundation/goteo/blob/devel/src/Goteo/Application/Event/ProjectValidationEvent.php).

This validation class can be captured (and changed) as shown (commented because the default behaviour is ok) in [ProjectListener](https://github.com/GoteoFoundation/goteo/blob/devel/src/Goteo/Application/EventListener/ProjectListener.php#L220).

By the way @alexm , this new implementation uses your Config variable (which is set to 3 by default). So it would be totally compatible with your solution.

Closing then this pull request. Thanks everyone!

PR #99, opened 2018-05-29
Fix table name worthcracy

alexm:
Trying to set a pending translation in /admin/worth raises the
following error:

    NO se ha marcado como pendiente de traducir!

References: https://github.com/UPC/goteo-nexus24/issues/9#issuecomment-391312944


alexm:
It seems that using the right table name isn't enough to set the pending status for translations: I just found out that the call to `setPending` passes an empty id.

alexm:
Now it works!

microstudi:
Thanks!

PR #116, opened 2019-02-05
Update DOCKER.md

davidbeig:
Explain that to execute 'docker/up' it's necessary that the user is in the docker group.


microstudi:
I've update the same file with this and other issues as well. Thanks anyway!

PR #127, opened 2019-08-06
Translations and most labels to English

cyberience:
Primary full translation to English for Contract,Contact,Admin,FAQ,Text_groups and Labels


cyberience:
Looks like you might need to fix the travis stuff because php7 changes. Not in my skill set, so you might need to figure a way how to import the translations, and probably is a reason there is out of date translations with this project. Good Luck

microstudi:
Thanks for this contribution, we are reviewing the copies.
It's best however to do that via [crowdin](https://translate.goteo.org).

PR #128, opened 2019-09-04
Override method idealiza inside User class

alexm:
A new call to idealiza with default values was added after merging https://github.com/GoteoFoundation/goteo/pull/97 disabling dots asvalid characters for username in that call (see commit f13c844d3).

This change will make the default value of `$punto` to be true for all existing and future calls inside User class.

### DIFF
--- a/src/Goteo/Model/User.php
+++ b/src/Goteo/Model/User.php
@@ -70,6 +70,11 @@ class User extends \Goteo\Core\Model {
     $webs = array(),
     $roles = array();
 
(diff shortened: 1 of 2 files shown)


alexm:
My change affects assertion in `tests/Goteo/Model/UserTest.php:96`:

```
$this->assertEquals('a-n', User::idealiza("a.ñ"));
```

I'll work on a fix.

microstudi:
Thanks Alex! we'll review & incorporate this in the next version!

PR #139, opened 2019-10-10
Update text foil

davidbeig:
Add function in Text to translate a copy to whatever lang we choose. Also, add that function to the list of functions provided in the extension of foil called TextUtils. This way in the templates there's the possibility of translating a copy to other langs and not only the one in the Lang::current(). This way communications from inside the platform can be send in different langs without having to change the session lang.

### DIFF
--- a/src/Goteo/Library/Text.php
+++ b/src/Goteo/Library/Text.php
@@ -57,28 +57,6 @@ class Text {
     }
 
     /**
-     * Compatibility text translation getter (instead of Lang::trans())
-     * @param  string $id key to translate (will be searched in all group translations)
-     * @param  [type] $lang [description]
-     * @return string     [description]
-     */
-    static public function getByLang ($id, $lang = null, $args = null) {
-        // Compatibilize with random number of arguments included
-        if(!is_array($args)) {
-            $args = func_get_args();
-            if (count($args) > 1) {
-                array_shift($args);
-            } else {
-                $args = array();
-            }
-        }
-
-        if (!$lang) $lang = Lang::current();
-
-        return static::lang($id, $lang, $args);
-    }
-
-    /**
      * Gets the required text.
      * If text does not exists a fallback will be returned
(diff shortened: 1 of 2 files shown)


src/Goteo/Util/Foil/Extension/TextUtils.php
```diff
@@ -45,7 +45,8 @@ public function provideFunctions()
           'sanitize' => [$this, 'sanitize'],
           'markdown' => [$this, 'markdown'],
           'to_rgba' => [$this, 'to_rgba'],
-          'image_src' => [$this, 'image_src']
+          'image_src' => [$this, 'image_src'],
+          'lang' => [$this, 'lang']
```
microstudi: Ca we just reuse the `t` function?

src/Goteo/Util/Foil/Extension/TextUtils.php
```diff
@@ -119,6 +120,10 @@ public function image_src($img, $w, $h, $ops = []) {
       return $src;
     }
 
+    public function lang() {
+        return call_user_func_array ( 'Goteo\Library\Text::getByLang' , func_get_args() );
```
microstudi: You can call `Goteo\Library\Text::lang` directly

src/Goteo/Library/Text.php
```diff
@@ -56,6 +56,28 @@ static public function sys($id, $args = null) {
...
+                array_shift($args);
+            } else {
+                $args = array();
+            }
+        }
+
+        if (!$lang) $lang = Lang::current();
+
+        return static::lang($id, $lang, $args);
+    }
```
microstudi: This whole thing about `func_get_args` is to maintain compatibility with the old behaviour, we don't want to use that in the future. Any use of this function should be substituted directly by `Text::lang`.
So I'd suggest to remove this function

microstudi (changes requested):

PR #164, opened 2021-04-12
Jgalobart home values module

jgalobart:
Add Goteo values module to home.

- tested in chrome and safari, desktop & mobile versions
- using same image for tabs but different files to provide different colours for each tab (not defined at design)
- tint images for footprint projects

Test:
- if partials are correct
- if JS compiles 
- if SASS compiles
- if footprint tabs works
- if slick slider works


jgalobart:
Project modules won't fit for large titles.

PR #171, opened 2021-04-26
Bug/fix unit tests

carlos-mg89:
Improved tests from this situation:

`Tests: 525, Assertions: 164741, Errors: 4, Failures: 3, Skipped: 14.`

To this one:

`OK (525 tests, 164838 assertions)`


microstudi:
@davidbeig step in!

composer.json
```diff
@@ -1,5 +1,6 @@
 {
     "require": {
+        "php": "^7.1",
```
carlos-mg89: At the Docker machine, we're working on PHP 7.1. I have no knowledge of what's the minimum PHP version on production, but if it's 7.1, I'd suggest to leave this in composer, so we are more aware of the current minimum version. Since PHPStorm was telling me 7.0 as minimum.

composer.json
```diff
@@ -47,7 +48,8 @@
         "sensiolabs/ansi-to-html": "^1.1",
         "unreal4u/telegram-api": "~3.4",
         "spipu/html2pdf": "^5.2",
-        "snowplow/referer-parser": "0.2.0"
+        "snowplow/referer-parser": "0.2.0",
+        "ext-pdo": "*"
```
carlos-mg89, line 52: Just wanna point out that PDO is required.

src/Goteo/Model/Sponsor.php
```diff
@@ -29,12 +28,8 @@ class Sponsor extends \Goteo\Core\Model {
         $name,
         $url,
         $image,
-        $type,
```
davidbeig, line 32: This field should not be removed. It is used in the Home of Goteo to filter the different types of Sponsors.

We should fix the test instead of removing it from the Model.

carlos-mg89: But there's no 'type' column in the database, neither in the 'goteo' DB nor in 'goteo_test'. Is it a field that is calculated somewhere else?

composer.json
davidbeig: Currently the production server is using PHP 7.0. But we are all working with the docker images with the 7.1. Maybe we can wait until we upgrade to 7.1 before setting it in the composer.json.

carlos-mg89: Definitely, will make it PHP 7.0 then.

src/Goteo/Model/Sponsor.php
davidbeig: There should be, we added it in this migration -> 20210317154642_goteo_add_sponsor_types.php

Can you check with the console that every migrate has been executed? 

```
docker/exec bin/console migrate status
```

carlos-mg89: Alright, it was very outdated my setup. Even though I had it setup from scratch, by current schema version is: 20201201150349

I'm gonna update it, retest everything again, and reintroduce the type column.

davidbeig: Thanks!

carlos-mg89: Just fixed it!!! And the test as well ;)

carlos-mg89:
@davidbeig all tests have been finally fixed ;)

microstudi (approved):

davidbeig (approved):

carlos-mg89:
Do you need anything else from here @davidbeig so you can merge the PR?

PR #174, opened 2021-05-10
Update composer dependencies

carlos-mg89:
Updated the following dependencies:
- PHPUnit (to 6.5) => This implied changing the namespace for the PHPUnit tests
- PHPMailer (to latest version)
- Mockery (to latest compatible version, needs updating PHP for a more recent version)
- tpyo/amazon-s3-php-class (with latest commit and fix version to it)
- Added required PHP extensions: pdo and openssl

Reduced verbosity while executing all tests, so we can focus on seeing the result of the tests and not specific errors.

More info: https://docs.google.com/document/d/1cfJqQ1Wl_15r8iQFrRNKNmdI8zPJcQrN98oi8Kzafrc/edit?usp=sharing

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -3,7 +3,7 @@
         "php": "^7.0",
         "foil/foil": "~0.6",
         "tpyo/amazon-s3-php-class": "~0.5",
-        "phpmailer/phpmailer": "~5.2",
+        "phpmailer/phpmailer": "^6.0",
         "jdorn/file-system-cache": "dev-master",
         "intervention/image": "~2.0",
         "mibe/feedwriter": "~1.0",
--- a/composer.json
+++ b/composer.json
@@ -49,7 +49,8 @@
         "unreal4u/telegram-api": "~3.4",
         "spipu/html2pdf": "^5.2",
         "snowplow/referer-parser": "0.2.0",
-        "ext-pdo": "*"
+        "ext-pdo": "*",
+        "ext-openssl": "*"
     },
     "autoload": {
         "psr-4": {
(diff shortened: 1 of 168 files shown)


composer.json
```diff
@@ -31,7 +31,7 @@
         "gregwar/captcha": "^1.1",
         "geoip2/geoip2": "~2.0",
         "openclerk/country-list": "^1.0",
-        "mockery/mockery": "^0.9.4",
+        "mockery/mockery": "^1.0",
```
davidbeig, line 36: Hi @carlos-mg89, I believe that in mockery 1.4 it drops support for PHP < 7.3, so I think this should be >= 1.0 and < 1.4. Maybe I'm missing something.

Thanks!

carlos-mg89: Yes, that's right. I've pointed this out in my notes (GoogleDrive). However, Composer won't let you update if PHP isn't 7.3. You can try by running composer update.

But yes, we can certainly restrict this in the composer.json. I'd do it right away ;)

davidbeig: True, I was missing that part!

carlos-mg89:
@davidbeig I've finally included in this PR the upgrade from PHPUnit 5.7 to 6.5

The huge part of the changes are due to the change on the PHPUnit namespace for the test classes.

And clearly, all tests continue on green ;)

davidbeig (approved):

davidbeig:
Thank you @carlos-mg89,

I have approved these changes. Really nice work! Later on I will merge them.

Seguim!

PR #180, opened 2021-06-08
Fixed PHP Docker container by changing it for one with Ubuntu 18.04 (…

carlos-mg89:
…instead of 16.04)

HOW TO RUN:

1. Be sure your Docker containers are down => `docker-compose down`
2. Perhaps you need to run this as well: `sudo rm -R .tmp/`
2. Remove your Docker Containers and build them again => `docker/up --build`


docker/php/Dockerfile
```diff
@@ -2,25 +2,39 @@
...
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* /usr/share/doc/*
 
+# Install Composer
+RUN php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');"
+RUN php -r "if (hash_file('sha384', 'composer-setup.php') === '756890a4488ce9024fc62c56153228907f1545c228516cbf63f885e036d37e9a59d27d63f46af1d4d07ee0f76181c7d3') { echo 'Installer verified'; } else { echo 'Installer corrupt'; unlink('composer-setup.php'); } echo PHP_EOL;"
+RUN php composer-setup.php
+RUN php -r "unlink('composer-setup.php');"
+RUN mv composer.phar /usr/local/bin/composer
+
```
davidbeig: Shouldn't we use something like this [How do I install Composer programmatically?](https://getcomposer.org/doc/faqs/how-to-install-composer-programmatically.md) ?

In case there's an update of composer, that will come eventually, this dockerfile won't work since the hash will be different.

docker/php/Dockerfile
```diff
@@ -2,25 +2,39 @@
...
     && apt-get -y --no-install-recommends install nodejs npm build-essential ruby-dev rubygems-integration gosu \
     && apt-get -y install zip unzip \
     && apt-get -y install mysql-client \
     && apt-get -y install jq \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* /usr/share/doc/*
 
+# Install Composer
+RUN php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');"
+RUN php -r "if (hash_file('sha384', 'composer-setup.php') === '756890a4488ce9024fc62c56153228907f1545c228516cbf63f885e036d37e9a59d27d63f46af1d4d07ee0f76181c7d3') { echo 'Installer verified'; } else { echo 'Installer corrupt'; unlink('composer-setup.php'); } echo PHP_EOL;"
```
davidbeig: RUN php -r 'copy("https://composer.github.io/installer.sig", "installer.sig");')"
RUN php -r "if (hash_file('sha384', 'composer-setup.php') === file_get_contents('installer.sig') { echo 'Installer verified'; } else { echo 'Installer corrupt'; unlink('composer-setup.php'); } echo PHP_EOL;"

davidbeig (commented):
This PR works great, I have set it up in my machine with no problems.

I just wrote two comments about the installation of composer. If you want I can apply those changes myself.

Thanks @carlos-mg89

docker/php/Dockerfile
carlos-mg89: Sounds very logical and your changes look good to me :)

davidbeig (approved):

microstudi (changes requested):
Sorry, this PR is base on the `beta` branch. It should be based on live. mergiing this as it is will mix beta into live.

Can you change @carlos-mg89 🙏

Great job on the other hand!

carlos-mg89:
Will don on Monday!

Missatge de Ivan Vergés ***@***.***> del dia dc., 9 de juny
2021 a les 17:07:

carlos-mg89:
@microstudi will do on Monday!

PR #181, opened 2021-06-14
Updated Dockerfile to use an Ubuntu image that's not deprecated

carlos-mg89:
HOW TO RUN:

1.  Be sure your Docker containers are down => `docker-compose down`
2.  Perhaps you need to run this as well => `sudo rm -R .tmp/`
3.  Remove your Docker Containers and build them again => `docker/up --build`

### DIFF
--- a/docker/php/Dockerfile
+++ b/docker/php/Dockerfile
@@ -2,28 +2,45 @@
 # PHPDocker.io PHP 7.1 / FPM image #
 ####################################
 
-FROM phpdockerio/php71-cli
+FROM ubuntu:bionic-20210512
 WORKDIR /application
+ARG PHP_VERSION=7.1
+
+# Install basic packages
+RUN apt-get update
+RUN apt-get -y install apt-utils software-properties-common wget apt-utils curl zip unzip mysql-client jq
+
+# Add Ondrej APT repository to have install several PHP versions
+RUN rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* /usr/share/doc/*
+RUN add-apt-repository ppa:ondrej/php
+RUN apt-key update
 
 # Install FPM
 RUN export DEBIAN_FRONTEND=noninteractive \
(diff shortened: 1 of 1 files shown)


carlos-mg89:
Useful info:

How to upgrade PHP version

- Edit `docker/php/Dockerfile` and change the `PHP_VERSION` variable
- Edit `docker-compose.yml` so we use the right PHP-FPM version.
  - E.g.: `./docker/php/php-ini-overrides.ini:/etc/php/7.1/fpm/conf.d/99-overrides.ini`
- Edit `docker/php/php-fpm-startup` to use the right PHP-FPM version
- Edit `grunt/run.js` so the first value on the `phps` array is the one you chose
- Edit `composer.json` so the min PHP version is the one you want to upgrade to

docker/php/Dockerfile
```diff
@@ -2,28 +2,45 @@
 # PHPDocker.io PHP 7.1 / FPM image #
 ####################################
 
-FROM phpdockerio/php71-cli
+FROM ubuntu:bionic-20210512
```
davidbeig: Could we change this to: 
```FROM ubuntu:bionic```?

I don't know for how long ubuntu keeps its images based on the compilation date

davidbeig (commented):
Thanks @carlos-mg89.

Great work here! I just left you one comment, and then we can merge!

docker/php/Dockerfile
carlos-mg89: @davidbeig done! Good advice, didn't notice about this, but I'm sure it's important. I've tested it with ubuntu:bionic and it works OK.

After reading this documentation https://hub.docker.com/_/ubuntu?tab=description&page=1&ordering=last_updated it seems there are more possibilities, like ubuntu:18.04 but I think we're good with this one :)

davidbeig (approved):
Perfect!

PR #182, opened 2021-06-14
Update composer dependencies

carlos-mg89:
Updated the following dependencies:

- PHPUnit (to 6.5) => This implied changing the namespace for the PHPUnit tests
- PHPMailer (to latest version)
- Mockery (to latest compatible version, needs updating PHP for a more recent version)
- tpyo/amazon-s3-php-class (with latest commit and fix version to it)
- Added required PHP extensions: pdo and openssl

Reduced verbosity while executing all tests, so we can focus on seeing the result of the tests and not specific errors.

More info: https://docs.google.com/document/d/1cfJqQ1Wl_15r8iQFrRNKNmdI8zPJcQrN98oi8Kzafrc/edit?usp=sharing


carlos-mg89:
Created new PR that replaces the approved one https://github.com/GoteoFoundation/goteo/pull/174 which was for devel branch

tests/Goteo/Core/ModelTest.php
```diff
@@ -133,8 +136,8 @@ public function testCreateTable($mock) {
                   `name` VARCHAR(100),
                   PRIMARY KEY (`id`),
                   UNIQUE INDEX (`uniq`) )";
-        // echo $sql;
-        $query = $mock::query($sql);
+        $mock::query($sql);
+        $this->assertTrue(true);
```
microstudi, line 140: and this? is there something that depends on this assertion?

microstudi (approved):
Super great @carlos-mg89 , just a silly thing!

tests/Goteo/Core/ModelTest.php
carlos-mg89: Yes, we prevent a warning from PHPUnit. Since I haven't found anything else to assert here, and I wanted to get rid of the warning (from PHPUnit 6 onward, a test must have at least one assert), I dealt with this like that.

It's not the best way to solve it, but it doesn't change the current test at all and there's no warning.

microstudi: super, merging then

PR #183, opened 2021-06-15
lock composer to symfony 2.8.30

carlos-mg89:
- Fix form saving due to issue with Symfony 2.8.52
- add composer.lock to git
- log to stdout if LOG_TO_STDOUT is present
- !!!!!!! Restore and early die if misconfiguration is found in settings (or database connection fails)


javicarrillo (approved):

carlos-mg89:
Looks perfect! I've been able to redeploy this Docker container, run all tests and verify it keeps working as it was.

PR #189, opened 2021-07-06
Use of env() parameters in configuration files

### LINKED ISSUE #188: Improve `YamlSettingLoader`
- [ ] Add a processing step after loading a file to extract matches like `%env(SOMETHING)%` and change its value ot the same corresponding ENV var specified in the environment (`getenv("SOMETHING")`)
image

We don't need the extra processing capabilites present in new versions of symfony (but we want to be compatible for the future)

davidbeig:
This PR enables our Config to handle values as environments.

When in the settings.yml there is an env() value, the Config class will search for that value using getenv().


src/Goteo/Application/Config.php
```diff
@@ -448,6 +448,11 @@ static public function get($name, $strict = false) {
 					$ret = null;
 				}
 			}
+
+			$pattern = "/^env\('(.*)'\)$/";
+			if (!is_array($ret) && preg_match($pattern, $ret, $env_regex)) {
+				$ret = getenv($env_regex[1]);
+			}
```
microstudi: this is ok, but we should move it to the `yamlsettingsloader` and then assume that Config has all the values already parsed

carlos-mg89: As far as I can see [here](https://symfony.com/doc/current/configuration.html#config-env-vars), the way of setting up variables in Symfony projects is this one:

```
framework:
    router:
        http_port: '%env(int:HTTP_PORT)%'
```

So I'd vote for implementing changes in a way that is compatible with Symfony.

I think the regular expression should be this one: `"/^%env\('(.*)'\)%$/"`

Anyway, I've been testing the above PR, and if I add an env parameter in my `local-docker-settings.yml`, I am given the name of the parameter instead of the value of the ENV parameter written in the .env file.

This is the env variable I've written in my YML file:
```
analytics:
    google: 'env(GOOGLE_ANALYTICS_ID)'
```

I'd expected to get in the result HTML the content of the GOOGLE_ANALYTICS_ID env parameter, but Instead I'm seeing this:

```
<script type="text/javascript">
    // @license magnet:?xt=urn:btih:0b31508aeb0634b347b8270c7bee4d411b5d4109&dn=agpl-3.0.txt

  (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
  (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
  m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
  })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');

    // Unamed first tracker
    ga('create', 'env(GOOGLE_ANALYTICS_ID)', 'auto');
    ga('send', 'pageview');


    // @license-end
</script>
```

Perhaps I understood a different desired output. But in the Symfony projects I've been using ENV variables, I've used them as I've described above. We have a unique .yml file with all the secrets, and the ones that need to use env parameters (because they change in different environments), we write them like this:

```
# config/packages/doctrine.yaml
doctrine:
    dbal:
        # by convention the env var names are always uppercase
        url: '%env(name_of_env_var)%'
```

microstudi: @carlos-mg89 you are right. And we need a test for this before merging.
So, do you agree, this is needed?:
- [ ] Process in the YamlSettingsLoader class instead of Config
- [ ] Add a test to ensure some cases and Standar Symfony Syntax

src/Goteo/Application/Config.php
```diff
@@ -423,36 +399,46 @@ static public function setConstants() {
...
+                } else {
+                    $paramValue = null;
+                }
+            }
+
+            if (!is_array($paramValue) && preg_match(self::ENV_PARAMETER_REG_EX, $paramValue, $matches)) {
+                if (sizeof($matches) >= 1) {
+                    $paramValue = getenv($matches[1]);
+                }
+            }
```
carlos-mg89, line 434: @microstudi I've finally approached this feature using the `Config` class.

I tried refactoring `YamlSettingsLoader` , but after some time trying to apply the read of ENV parameters there, looked to me as changing too much that class.

Keeping the parsing of ENV parameters in `Config` will allow us having any kind of config file, not just Yaml as well.

To test this, you can simply do this:
1. Edit the `local-docker-settings.yml` to include an ENV variable. E.g.: `db.port: '%env(DATABASE_PORT)%'`
2. Edit the `docker-compose.yml` to have that ENV variable:
```
environment:
- DATABASE_PORT=3306
```

You'd have to run `docker-compose down` and `docker/build` in order to get the ENV variables loaded. I tried by editing the `.env` file instead of using Docker, but it just did't work :(

microstudi (approved):
Thanks a lot @carlos-mg89 this is what we need!
Next step reorganize settings.yml and leave only one file!!

src/Goteo/Application/Config.php
davidbeig: > Anyway, I've been testing the above PR, and if I add an env parameter in my `local-docker-settings.yml`, I am given the name of the parameter instead of the value of the ENV parameter written in the .env file.
> 
> This is the env variable I've written in my YML file:
> 
> ```
> analytics:
>     google: 'env(GOOGLE_ANALYTICS_ID)'
> ```

I think this is because the regex was expecting something like env('name_of_env_var'), I guess it was not followint the correct syntax like now with your changes.

carlos-mg89: Yes @davidbeig , I realized about the single quotes when I added the % characters. You did a very good job with the initial PR by the way!

davidbeig: We all did a good job!

PR #212, opened 2021-09-06
Feature/home values dynamic

davidbeig:
## What?

A module in the home of Goteo to show the values of the platform. These values are related to the footprints, the sdgs and the projects.

## Why?

To be able to explain the impact of Goteo, the projects that are being funded and the framework of footprints that Goteo created.

## How?

Creating a partial that can be used in the home of Goteo.
To use the impact_data model there is a migration to create a footprint_impact table that can be used through the ImpactDataRelationshipTrait

## Anything Else?

This PR is related to #213. First we have to merge that one.

This PR is based on the work made by @jgalobart and integrating it completely with the logic inside Goteo.


davidbeig:
- [ ] Update migration datetime

composer.json
```diff
@@ -11,25 +11,25 @@
...
+        "symfony/console": "2.8.30",
+        "symfony/translation": "2.8.30",
+        "symfony/finder": "2.8.30",
+        "symfony/dom-crawler": "2.8.30",
+        "symfony/css-selector": "2.8.30",
+        "symfony/process": "2.8.30",
+        "symfony/form": "2.8.30",
+        "symfony/templating": "2.8.30",
+        "symfony/framework-bundle": "2.8.30",
+        "symfony/validator": "2.8.30",
```
carlos-mg89, line 32: Beware with this change, since in production we already have 2.8.52

PR #213, opened 2021-09-07
Feature impact data

davidbeig:
## What?

We have created an Impact Data model that is meant to be used by other classes using the ImpactDataRelationshipTrait.

## Why?

To be able to create data for every model about their impact

## How?

Creating a table to store data related to impact and a Trait that can be used by other classes.

## Anything Else?


src/Goteo/Controller/Admin/ImpactDataAdminController.php
```diff
@@ -0,0 +1,110 @@
...
+{
+    protected static $icon = '<i class="fa fa-2x fa-table"></i>';
+
+    public static function getGroup() {
+        return 'communications';
+    }
+
+    public static function getRoutes()
+    {
+        return [
```
javicarrillo, line 31: Add translate action

javicarrillo (changes requested):
- Rename migration to current date
- Add translate action

src/Goteo/Library/Forms/Model/ImpactDataForm.php
```diff
@@ -0,0 +1,105 @@
...
+
+    public function createForm() {
+
+        $model = $this->getModel();
+        $builder = $this->getBuilder();
+        $options = $builder->getOptions();
+        $data = $options['data'];
+
+        $builder
+            ->add('title', 'text', array(
```
carlos-mg89: @davidbeig would you mind using the new Symfony Form syntax instead?

So instead of 'text' it would be TextType::class (you'd need to import the class as well). 

That way, we would have less work to do when we merge the Symfony 3.4 branch into the live branch. Which could happen before or after this one.

The same for the other types in this form.

src/Goteo/Util/ModelNormalizer/Transformer/ImpactDataTransformer.php
```diff
@@ -0,0 +1,33 @@
...
+ * and LICENSE files that was distributed with this source code.
+ */
+namespace Goteo\Util\ModelNormalizer\Transformer;
+
+use Goteo\Core\Model;
+use Goteo\Library\Text;
+
+/**
+ * Transform a Model
+ */
```
carlos-mg89: In my opinion, the class name is quite self explanatory and this comment is not adding value. If you agree, I'd vote to simply remove the comment.

public/index.php
```diff
@@ -48,6 +48,17 @@
 // Get the app
 $app = App::get();
 
+if(getenv('LOG_TO_STDOUT')) {
```
javicarrillo: This file shouldn't be here, it is updated in live.

javicarrillo (changes requested):

javicarrillo (approved):

PR #218, opened 2021-09-07
Update live base code with latest fixes

davidbeig:
This PR brings the lasts developments to the live branch.


public/index.php
```diff
@@ -48,6 +48,17 @@
 // Get the app
 $app = App::get();
 
+if(getenv('LOG_TO_STDOUT')) {
```
microstudi: this change is not needed anymore, this is done in a plugin now

microstudi (changes requested):
Pleas add a meaninful title, remove changes index changes an will merge

microstudi (approved):
thanks!

Issue #223, opened 2021-09-08
Remove grunt cli version  1.3.2 on Dockerfile


davidbeig:
Since the upgrade of the Dockerifle images we are using, and with the corresponding improvement, we don't have to limit the version of grunt-cli that we install to @1.3.2 because of the nodejs version that is being installed in the ubuntu distribution we are using.

This Issue is here to revert #165.

PR #224, opened 2021-09-08
Upgrade Grunt and grunt-contrib-uglify

davidbeig:
Upgrade grunt inside our Dockerfile images since we don't have the limitation of the old ubuntu 16.04 distribution. Since with the Ubuntu 18.04 image we have node 8.0 we can upgrade the Grunt version we are using.

This also allows us to upgrade the uglify minifier that we were using so that it understands ES6.

This PR is related to #223

### DIFF
--- a/docker/php/Dockerfile
+++ b/docker/php/Dockerfile
@@ -36,7 +36,7 @@ RUN mv composer.phar /usr/local/bin/composer
 
 RUN gem install sass -v 3.4.23
 RUN gem install compass
-RUN npm install -g grunt-cli@1.3.2
+RUN npm install -g grunt-cli
 
 # PHP-FPM packages need a nudge to make them docker-friendly
 COPY overrides.conf /etc/php/${PHP_VERSION}/fpm/pool.d/z-overrides.conf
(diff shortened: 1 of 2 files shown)


docker/php/Dockerfile
```diff
@@ -36,7 +36,7 @@ RUN mv composer.phar /usr/local/bin/composer
 
 RUN gem install sass -v 3.4.23
 RUN gem install compass
-RUN npm install -g grunt-cli@1.3.2
+RUN npm install -g grunt-cli
```
microstudi:
```suggestion
RUN npm install -g grunt-cli@4.0.0
```
Shouldn't we lock, just in case?

davidbeig: The v4.0.0 version is the one of grunt-contrib-uglify. We can lock the version of grunt-cli to v1.4.1, since the 1.4.2 bumps the minimum version of node to v10.

davidbeig: Also, maybe we can move this dependency from the Dockerfile to the package.json?

microstudi: we need it globally as it is used as installed program.
I'd require here the exact same version specified in package.json

microstudi (approved):
thanks!

PR #234, opened 2021-09-15
Feature: Project can share a url to a signature platform

davidbeig:
# What
 Allow projects to show a url to a signature platform.

# Why
 Some projects that use Goteo as crowdfunding platform have a parallel campaign where they try to get people to sign a petition, whether it is for a citizens initiative, a political campaing, etc

# How
 Add to the project model the necessary fields that can be used to share the signature platform inside the project layout.


javicarrillo (approved):
Great Job!

PR #254, opened 2021-10-15
Legal Entities and Legal Document Types model

davidbeig:
- add LegalEntity model and subclasses
- add LegalDocumentType model and subclasses


src/Goteo/Model/LegalEntity.php
```diff
@@ -0,0 +1,53 @@
...
+            case self::LEGAL_PERSON:
+                $legal_entity_type = new LegalPerson();
+            case self::NATURAL_PERSON:
+                $legal_entity_type = new NaturalPerson();
+            default:
+                $legal_entity_type = new NaturalPerson();
+        }
+
+        return $legal_entity_type;
+    }
```
carlos-mg89, line 26: The above method could be simplified to:

```
public static function create(string $legal_entity): LegalEntity {
    switch ($legal_entity) {
        case self::LEGAL_PERSON:
            return new LegalPerson();
        default:
            return new NaturalPerson();
    }
}
```

src/Goteo/Model/LegalDocumentType/NIE.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace Goteo\Model\LegalDocumenType;
+
+class NIE {
```
carlos-mg89: All these classes (under the LegalDocumentType folder), shouldn't extend from `LegalDocumentType`?

Otherwise they aren't really a `LegalDocumentType`. Plus, they won't have access to the `self::NIE` constant either.

src/Goteo/Model/LegalDocumentType/NIE.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace Goteo\Model\LegalDocumenType;
```
carlos-mg89: The namespace should be `LegalDocumentType`, not `LegalDocumenType`. Same applies to the rest of the classes in this namespace. All have the typo.

src/Goteo/Model/LegalDocumentType.php
```diff
@@ -0,0 +1,59 @@
...
+namespace Goteo\Model;
+
+use Goteo\Library\Text;
+
+use Goteo\Core\Model;
+use Goteo\Model\LegalDocument\NIE;
+use Goteo\Model\LegalDocument\CIF;
+use Goteo\Model\LegalDocument\NIF;
+use Goteo\Model\LegalDocument\VAT;
+use Goteo\Model\LegalDocument\PASSPORT;
```
carlos-mg89: Shouldn't the imports be like this?

```
use Goteo\Model\LegalDocumentType\CIF;
use Goteo\Model\LegalDocumentType\NIE;
use Goteo\Model\LegalDocumentType\NIF;
use Goteo\Model\LegalDocumentType\PASSPORT;
use Goteo\Model\LegalDocumentType\VAT;
```

src/Goteo/Model/LegalDocumentType/CIF.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace Goteo\Model\LegalDocumentType;
+
+class CIF extends LegalDocumentType {
```
carlos-mg89, line 7: It's missing the import of the extended class:

`use Goteo\Model\LegalDocumentType;`

src/Goteo/Model/LegalDocumentType.php
```diff
@@ -0,0 +1,56 @@
...
+    const CIF = 'cif';
+    const NIF = 'nif';
+    const NIE = 'nie';
+    const VAT = 'vat';
+    const PASSPORT = 'passport';
+    const LEGAL_DOCUMENTS = [self::CIF, self::NIF, self::NIE];
+
+    private string $document_type;
+
+    public static function create(string $document_type): LegalDocument {
```
carlos-mg89: @davidbeig I think that the return type should be `LegalDocumentType`. Isn't it?

src/Goteo/Model/LegalEntity.php
```diff
@@ -0,0 +1,47 @@
...
+            default:
+                return new NaturalPerson();
+        }
+    }
+
+    public function getEntity(): string {
+        return $this->legal_entity;
+    }
+
+    public static function getNaturalPerson(): string {
```
carlos-mg89: I think that we could use the constant directly and remove this method

src/Goteo/Model/LegalEntity.php
```diff
@@ -0,0 +1,47 @@
...
+        return self::NATURAL_PERSON;
+    }
+
+    public static function getLegalPerson(): string {
```
carlos-mg89: I think that we could use the constant directly and remove this method

PR #268, opened 2021-11-04
Added ConsoleExceptionListener as an EventSubscriber

carlos-mg89:


src/container.php
```diff
@@ -161,6 +161,7 @@
   ->addMethodCall('addSubscriber', array(new Reference('app.listener.channel')))
   ->addMethodCall('addSubscriber', array(new Reference('console.listener.milestone')))
   ->addMethodCall('addSubscriber', array(new Reference('console.listener.favourite')))
+  ->addMethodCall('addSubscriber', array(new Reference('console.listener.exception')))
```
carlos-mg89, line 164: It looks like is not enough to register the console.listener.exception, but that we must add it to the `addSubscriber` method of the `Symfony\Component\EventDispatcher\EventDispatcher`

src/Goteo/Console/Console.php
```diff
@@ -35,9 +35,9 @@ static public function get() {
 
 			self::$app = App::getService('console');
             self::$console = new Application();
-            /** @var EventDispatcherInterface $dispatcher */
-            $dispatcher = App::getService('dispatcher');
-            self::$console->setDispatcher($dispatcher);
+            /** @var EventDispatcherInterface $consoleDispatcher */
+            $consoleDispatcher = App::getService('console_dispatcher');
+            self::$console->setDispatcher($consoleDispatcher);
```
carlos-mg89: @davidbeig here it was the clue about why the `ConsoleDispatcher` wasn't handling the `ConsoleListeners`, alongside with the missing parameter in the `container.php`

src/container.php
```diff
@@ -146,24 +146,22 @@
...
-  ->addMethodCall('addSubscriber', array(new Reference('app.listener.invest')))
-  ->addMethodCall('addSubscriber', array(new Reference('app.listener.poolinvest')))
-  ->addMethodCall('addSubscriber', array(new Reference('app.listener.donateinvest')))
-  ->addMethodCall('addSubscriber', array(new Reference('app.listener.messages')))
-  ->addMethodCall('addSubscriber', array(new Reference('app.listener.project_post')))
-  ->addMethodCall('addSubscriber', array(new Reference('app.listener.blog_post')))
-  ->addMethodCall('addSubscriber', array(new Reference('app.listener.stories')))
-  ->addMethodCall('addSubscriber', array(new Reference('app.listener.channel')))
-  ->addMethodCall('addSubscriber', array(new Reference('console.listener.milestone')))
-  ->addMethodCall('addSubscriber', array(new Reference('console.listener.favourite')))
```
carlos-mg89, line 163: @davidbeig @microstudi any good reason why would we want to have `console.listener.milestone` and `console.listener.favourite` added both for the `EventDispatcher` registered as 'dispatcher' and for the `console_dispatcher` as well?

Could we have it only in the `console_dispatcher`?

src/Goteo/Console/Command/TestCommand.php
```diff
@@ -22,9 +22,9 @@ protected function configure()
     {
         $this->setName("test")
              ->setDescription("A mail testing command")
-             ->setDefinition(array(
-                      new InputOption('fail', 'f', InputOption::VALUE_NONE, 'Do a fail test and send a failure mail'),
-                ))
+             ->setDefinition([
+                 new InputOption('fail', 'f', InputOption::VALUE_NONE, 'Do a fail test and send a failure mail')
```
microstudi:
```suggestion
                 new InputOption('fail', 'f', InputOption::VALUE_NONE, 'Do a fail test and send a failure mail (if --logmail option is used)')
```

microstudi: I would improve the help here

src/Goteo/Console/Command/TestCommand.php
```diff
@@ -42,9 +42,10 @@ protected function configure()
      */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
-        if ( $input->getOption('fail') ) {
+        if ($input->getOption('fail')) {
            $output->writeln('<comment>Throwing some errors intentionally to test email sending</comment>');
```
microstudi:
```suggestion
           $output->writeln('<comment>Throwing some errors intentionally...</comment>');
```

src/Goteo/Console/Console.php
microstudi: this is because of this change 
image

src/container.php
davidbeig: Yes, per the ConsoleMilestoneListener, it listens to events defined in ConsoleEvents and in AppEvents. Some are thrown inside a console command, mostly executed using cron, and the others are thrown using the workflow of the application (In this case through and invest)

microstudi: Because the ConsoleMilestoneListener is used both in the web application and the console, events managed are defined here:
https://github.com/GoteoFoundation/goteo/blob/0a5fc684813287e011fcaa77e151d16a73dad002/src/Goteo/Console/EventListener/ConsoleMilestoneListener.php#L150-L157

Maybe it could be done in 2 different classes for clarity with some composition (a trait?) for sharing methods?

microstudi (commented):
Thanks @carlos-mg89 looks nice. We should tested in a staging env first

davidbeig (commented):
Great job finding it @carlos-mg89 ! Looks like we are still missing some tests for the commands... haha

Right now it throws a Command Error when executing bin/console mailing. 

```Oops, exception thrown while running command mailing
Command Exception
	error: Call to undefined method Goteo\Console\Console::getDispatcher()
	command: mailing
	options: Array
	trace: Error: Call to undefined method Goteo\Console\Console::getDispatcher()
```
I've seen that the Console.php has been also upgraded with the last Symfony upgrade.

I think we have to fix some things there.

src/container.php
carlos-mg89: For now, I'll simply put them back. Let's not delay the fix.

src/Goteo/Console/Console.php
carlos-mg89: Alright, however, I believe now is fine as well, isn't it? 

And perhaps it's even more clear which dispatcher in being used than with `self::$app->getDispatcher()`.

microstudi (approved):

PR #276, opened 2021-11-10
Include HttpClient when creating Payment Gateway (it seems none was added)

carlos-mg89:


src/Goteo/Payment/Method/AbstractPaymentMethod.php
```diff
@@ -22,39 +23,49 @@
...
+     * @throws PaymentException
+     * @throws Config\ConfigException
+     */
     private function initGateway()
     {
         $factory = new GatewayFactory();
-        $this->gateway = $factory->create($this->getGatewayName());
+        $this->gateway = $factory->create(
+            $this->getGatewayName(),
+//            new GuzzleClient() // @TODO Move GuzzleClient from goteo-private into goteo
```
carlos-mg89: It seems this is where the HttpClient is set. However, if we let the GatewayFactory with a null HttpClient, it's unclear what happens afterwards...

davidbeig (approved):
Thanks @carlos-mg89 for cleaning this up!

PR #285, opened 2021-11-17
Feature/home values module

davidbeig:


db/migrations/20180919153237_goteo_sdg.php
```diff
@@ -119,7 +119,15 @@ public function postUp()
...
-      if(!$foot->save($errors)) {
+      $fields = ['name', 'icon', 'description'];
+
+      try {
+          $foot->dbInsertUpdate($fields);
+      }
+      catch(\PDOException $e) {
+          $errors[] = 'Error saving footprint: ' . $e->getMessage();
+      }
+      if(!empty($errors)) {
```
carlos-mg89, line 130: @davidbeig What's this? Is this related with this PR?

public/assets/js/impact_discover/impact_discover.js
```diff
@@ -0,0 +1,261 @@
...
+section 4, provided you include this license notice and a URL
+through which recipients can access the Corresponding Source.
+
+
+@licend  The above is the entire license notice
+for the JavaScript code in this page.
+*/
+
+//reset ODS JSON
+function resetODS() {
```
carlos-mg89, line 27: @davidbeig would be nicer to rewrite these JS functions so they contain all the information that the comments have. So we can get rid of all these comments.

public/assets/js/impact_discover/impact_discover.js
```diff
@@ -0,0 +1,261 @@
...
+// reset ODS select
+function resetODSSelect() {
+    $(".impact-discover-filters select option").remove();
+    $(".impact-discover-filters select").append("<option>Filtra por Objetivos de Desarrollo sostenible</option");
+}
+
+// check if ODS option has selected footprint
+function hasFootprint (ods, footprint) {
+    var footprints = ods.footprints;
+    return footprints.indexOf(footprint)>=0 ? true : false;
```
carlos-mg89, line 42: @davidbeig Wouldn't this be the same than:
`return footprints.indexOf(footprint) >= 0;`

PR #286, opened 2021-11-18
Footprint impact data migration

davidbeig:
This PR has the migration necessary for the new module in the home developed in feature/home_values_module

### DIFF
--- a/db/migrations/20210930185746_goteo_footprint_impact_data.php
+++ b/db/migrations/20210930185746_goteo_footprint_impact_data.php
@@ -1,0 +1,63 @@
+<?php
+/**
+ * Migration Task class.
+ */
+class GoteoFootprintImpactData
+{
+  public function preUp()
+  {
+      // add the pre-migration code here
+  }
+
+  public function postUp()
+  {
+      // add the post-migration code here
+  }
+
+  public function preDown()
+  {
+      // add the pre-migration code here
+  }
+
+  public function postDown()
+  {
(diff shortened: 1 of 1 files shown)


db/migrations/20210930185746_goteo_footprint_impact_data.php
```diff
@@ -0,0 +1,63 @@
...
+  /**
+   * Return the SQL statements for the Up migration
+   *
+   * @return string The SQL string to execute for the Up migration.
+   */
+  public function getUpSQL()
+  {
+     return "
+      CREATE TABLE `footprint_impact` (
+        `footprint_id` INT(10) UNSIGNED NOT NULL,
```
microstudi, line 36: No id in this table?

davidbeig: This table is like the ones we have for sdg_footprint were it only has a relation between footprint and impact_data

carlos-mg89 (dismissed):

carlos-mg89:
@davidbeig I believe the reason why this PR test is failing is because it depends on https://github.com/GoteoFoundation/goteo/pull/283 being approved first. This migration believes the tabla `impact_data` exists, however it's created in that other PR.

Why don't we merge both PRs? I would move this migration to the other PR.

javicarrillo (dismissed):

javicarrillo (approved):

PR #298, opened 2021-11-23
Impact data model

davidbeig:
What?

We have created an Impact Data model that is meant to be used by other classes using the ImpactDataRelationshipTrait.

Why?

To be able to create data for every model about their impact
How?

Creating a table to store data related to impact and a Trait that can be used by other classes.
Testing?
Screenshots (optional)
Anything Else?


src/Goteo/Application/Config.php
```diff
@@ -336,6 +338,7 @@ static public function addAdminControllers()
         AdminController::addSubController(ChannelPostsAdminController::class);
         AdminController::addSubController(ChannelSectionAdminController::class);
         AdminController::addSubController(ChannelStoryAdminController::class);
+		AdminController::addSubController(ImpactDataAdminController::class);
```
carlos-mg89: Unaligned indentation

src/Goteo/Controller/Admin/ImpactDataAdminController.php
```diff
@@ -0,0 +1,110 @@
...
+use Goteo\Application\Message;
+use Goteo\Library\Text;
+use Goteo\Library\Forms\Model\ImpactDataForm;
+use Goteo\Model\ImpactData;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\Routing\Route;
+
+class ImpactDataAdminController extends AbstractAdminController
+{
+    protected static $icon = '<i class="fa fa-2x fa-table"></i>';
```
carlos-mg89, line 23: Why do we need HTML code in a PHP controller?

src/Goteo/Controller/Admin/ImpactDataAdminController.php
```diff
@@ -0,0 +1,110 @@
...
+        $list = ImpactData::getList([], $page * $limit, $limit, false);
+        $total = ImpactData::getList([], 0, 0, true);
+        
+        return $this->viewResponse('admin/impact_data/list',[
+            'list' => $list,
+            'total' => $total
+        ]);
+    }
+
+    public function editAction($id = '', Request $request)
```
carlos-mg89: The `$request` param should be first here. Parameters with predefined values should always be at the end (because they are optional).

src/Goteo/Library/Forms/Model/ImpactDataForm.php
```diff
@@ -0,0 +1,118 @@
...
+
+        $model = $this->getModel();
+        $builder = $this->getBuilder();
+        $options = $builder->getOptions();
+        $data = $options['data'];
+
+        $builder
+            ->add('title', TextType::class, array(
+                'label' => 'regular-title',
+                'required' => true,
```
carlos-mg89: By default, Symfony Form fields are required, so it's not needed to write this down: https://symfony.com/doc/current/forms.html#the-required-option

src/Goteo/Model/ImpactData.php
```diff
@@ -0,0 +1,128 @@
...
+        list($fields, $joins) = self::getLangsSQLJoins($lang);
+
+        $sql = "SELECT
+                    id,
+                    $fields,
+                    image,
+                    lang
+                FROM impact_data
+                $joins
+                WHERE id = :id";
```
carlos-mg89, line 48: For the next time that we need to create a new model, I would strongly suggest that we take the time to configure Doctrine ORM and we start using it for that model.

tests/Goteo/Model/ImpactDataTest.php
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    /**
+     * @depends testGetImpactData
+     */
+    public function testRemove(ImpactData $impact_data) {
+
+        $count = ImpactData::getList([],0,0,true);
+        $this->assertTrue($impact_data->dbDelete());
+        $this->assertCount($count - 1, ImpactData::getList());
```
carlos-mg89: It's not that there is only a valid way of writing tests, however there's a known pattern that advocates for separating the lines a the test in 3 blocks: GIVEN, WHEN and THEN. More info: https://martinfowler.com/bliki/GivenWhenThen.html

GIVEN these variables and/or objects
WHEN we execute this method
THEN these asserts will be true

Example with the above case:

```
$impactDataCount = ImpactData::getList([],0,0,true);

$response = $impactData->dbDelete();

$this->assertTrue($response);
$this->assertCount($impactDataCount - 1, ImpactData::getList());
```

src/Goteo/Controller/Admin/ImpactDataAdminController.php
davidbeig: Since @javicarrillo and @microstudi developed the new modules, they have this protected $icon that is used to show, normally, a font-awesome icon and it has been always added through that html code inside the variable. We could change it.

carlos-mg89: Cool, we could later rewrite this

tests/Goteo/Model/ImpactDataTest.php
```diff
@@ -87,10 +87,12 @@ public function testList() {
...
-        $count = ImpactData::getList([],0,0,true);
-        $this->assertTrue($impact_data->dbDelete());
-        $this->assertCount($count - 1, ImpactData::getList());
-    }
+        $impactDataCount = ImpactData::getList([],0,0,true);
+
+        $response = $impactData->dbDelete();
+
+        $this->assertTrue($response);
+        $this->assertCount($impactDataCount - 1, ImpactData::getList());    }
```
carlos-mg89: Cool! We're almost there. To finish with this, please, move the } to a new line.

src/Goteo/Library/Forms/Model/ImpactDataForm.php
```diff
@@ -35,27 +35,23 @@ public function createForm() {
         $builder
             ->add('title', TextType::class, array(
                 'label' => 'regular-title',
-                'required' => true,
```
carlos-mg89: Wonderful!

javicarrillo (approved):

PR #299, opened 2021-11-23
Feature Values Module

davidbeig:
This PR creates the values module that can be used inside the home of the application. It also uses the new impact data module and provides infor regarding projects and the related sdgs.


db/migrations/20180919153237_goteo_sdg.php
```diff
@@ -119,7 +119,15 @@ public function postUp()
...
-      if(!$foot->save($errors)) {
+      $fields = ['name', 'icon', 'description'];
+
+      try {
+          $foot->dbInsertUpdate($fields);
+      }
+      catch(\PDOException $e) {
+          $errors[] = 'Error saving footprint: ' . $e->getMessage();
+      }
+      if(!empty($errors)) {
```
carlos-mg89, line 130: @davidbeig What's this migration change? Is it really related with this PR?

public/assets/js/impact_discover/impact_discover.js
```diff
@@ -0,0 +1,261 @@
...
+section 4, provided you include this license notice and a URL
+through which recipients can access the Corresponding Source.
+
+
+@licend  The above is the entire license notice
+for the JavaScript code in this page.
+*/
+
+//reset ODS JSON
+function resetODS() {
```
carlos-mg89: @davidbeig would be nicer to rewrite these JS functions so they contain all the information that the comments have. So we can get rid of all these comments.

public/assets/js/impact_discover/impact_discover.js
```diff
@@ -0,0 +1,261 @@
...
+function removeODS(ods){
+    $.each(odsList.ods, function(key, value){
+        if (value.id == ods) odsList.ods[key]["active"] = false
+    })
+    refreshODS();
+}
+
+function isODSActive(ods) {
+    return ods.active;
+}
```
carlos-mg89: @davidbeig is this method really necessary? Couldn't we use ods.active in the other JS code?

public/assets/js/impact_discover/impact_discover.js
```diff
@@ -0,0 +1,261 @@
...
+function resetODSSelect() {
+    $(".impact-discover-filters select option").remove();
+    $(".impact-discover-filters select").append("<option>Filtra por Objetivos de Desarrollo sostenible</option");
+}
+
+// check if ODS option has selected footprint
+function hasFootprint (ods, footprint) {
+    var footprints = ods.footprints;
+    return footprints.indexOf(footprint)>=0 ? true : false;
+}
```
carlos-mg89: @davidbeig Wouldn't this be the same than:
`return ods.footprints.indexOf(footprint) >= 0;`

public/assets/js/impact_discover/impact_discover.js
```diff
@@ -0,0 +1,261 @@
...
+
+function needsScroll() {
+    var $last = $('.impact-discover-projects div:last');
+    return $(window).scrollTop() >= $last.offset().top - $last.height()
+}
+
+$(window).on('resize scroll', function() {
+    if(needsScroll() && !$('.loading-container').length) {
+        ods = [];
+        //TODO: find selected ODS
```
carlos-mg89: @davidbeig is this TODO comment needed?

public/assets/js/impact_discover/impact_discover.js
```diff
@@ -0,0 +1,261 @@
...
+});
+
+// bind click on footprint
+$("#filters-footprints").on("click","a", function(e){
+    e.preventDefault();
+    resetFootprints();
+    footprint = $(this).attr("data-footprint");
+    filterODSByFootprint(footprint);
+    $(this).addClass("active");
+    if ($(window).width()<721) {
```
carlos-mg89: @davidbeig why 721? Is this a value that could be set as a constant in a class so we can reuse it across the whole project?

src/Goteo/Controller/ImpactDiscoverController.php
```diff
@@ -0,0 +1,44 @@
...
+use Goteo\Model\Project\ProjectLocation;
+
+class ImpactDiscoverController extends \Goteo\Core\Controller {
+
+    public function __construct() {
+        // Cache & replica read activated in this controller
+        $this->dbReplica(true);
+        $this->dbCache(true);
+
+        // \Goteo\Core\DB::replica(true);
```
carlos-mg89: @davidbeig I'm pretty sure that this line can be removed

src/Goteo/Controller/ImpactDiscoverController.php
```diff
@@ -0,0 +1,44 @@
...
+use Goteo\Application\View;
+use Goteo\Application\Session;
+use Goteo\Model\Category;
+use Goteo\Model\Project;
+use Goteo\Model\Project\ProjectLocation;
+
+class ImpactDiscoverController extends \Goteo\Core\Controller {
+
+    public function __construct() {
+        // Cache & replica read activated in this controller
```
carlos-mg89, line 29: @davidbeig I'm pretty sure that this line can be removed

src/routes.php
```diff
@@ -267,4 +267,17 @@
     )
 ));
 
+// $map_routes = include __DIR__ . '/Routes/map_routes.php';
+// $map_routes->addPrefix('/map');
+// $routes->addCollection($map_routes);
```
carlos-mg89, line 272: @davidbeig could these commented lines be removed?

db/migrations/20180919153237_goteo_sdg.php
davidbeig: Since the new migration for this two modules (the one that creates title for footprint), this migration is broken, that's why we create those footprints only using the name, icon and description. I guess that this commit is also inside the other PRs...

public/assets/js/impact_discover/impact_discover.js
davidbeig: This is an outdated code.  Right now is as follows:
```
    function hasFootprint (sdg, footprint) {
        var footprints = sdg.footprints;
        return footprints.find(element => element.id == footprint) != undefined;
    }
```

public/assets/js/impact_discover/impact_discover.js
davidbeig: True, this value was chosen by Jordi, inside our scss you can find the definitions we have for the different breakpoints inside _variables.scss. In this case we define it like this

```
$breakpoint-sm: 768px;
```

We should really import this defined variables to our javascript

src/routes.php
davidbeig: We could also move the previously defined routes inside this map_routes.php.

javicarrillo (approved):

PR #300, opened 2021-11-23
Feature impact discover

davidbeig:
This PR creates the new impact discover landing.


Resources/templates/responsive/impact_discover/partials/projects_third_row.php
```diff
@@ -0,0 +1,26 @@
...
+                </div>
+            <?php endif; ?>
+        </div>
+    </div>
+    <div href="/project/<?= current($projects)->id ?>" class="col col-xs-12 col-sm-8">
+        <?php if(next($projects)): ?>
+            <?= $this->insert('impact_discover/partials/project.php', ['project' => current($projects)]) ?>
+        <?php endif; ?>
+    </div>
+</div>
```
carlos-mg89: @davidbeig couldn't these 3 templates be handled in just one template?

src/Goteo/Application/Config.php
```diff
@@ -183,6 +183,9 @@ static public function load($config_file) {
 			View::addFolder(__DIR__ . '/../../../Resources/templates/responsive');
             View::setTheme('responsive');
 
+            \Goteo\Application\View::setTheme('responsive');
+            // we die here and show a formatted error, most likely reason is a database misconfiguration
+			die(\Goteo\Application\View::render('errors/config', ['msg' => $e->getMessage(), 'info' => $info, 'file' => $config_file, 'code' => 500], false));
```
carlos-mg89: @davidbeig wouldn't this be an issue in production?

src/Goteo/Controller/Api/ProjectsApiController.php
```diff
@@ -625,4 +627,88 @@ public function projectDeleteSupportMsgAction($mid) {
...
+                'page' => $page,
+                'limit' => $limit,
+                'result_total' => count($projects),
+                'html' => View::render( 
+                    'impact_discover/partials/list_project_rows', [
+                        'projects' => $projects
+                    ]
+                )
+            ];
+        }
```
carlos-mg89: @davidbeig I think we could have a switch instead of if / else if here. Followed by calling 3 different methods that return the response.

It would be much simpler to read and to understand that way.

src/Goteo/Application/Config.php
davidbeig: True, i think this one has slipped when merging with the current state of live

src/routes.php
```diff
@@ -267,4 +267,51 @@
...
+        '_controller' => 'Goteo\Controller\ImpactDiscoverController::mapAction'
+    )
+));
+
+$routes->add('impact-discover-mosaic', new Route(
+    '/impact-discover/mosaic',
+    array(
+        '_controller' => 'Goteo\Controller\ImpactDiscoverController::mosaicAction'
+    )
+));
```
carlos-mg89: @davidbeig aren't the routes being added twice here?

Resources/templates/responsive/impact_discover/partials/projects_third_row.php
davidbeig: Yes, they could, i just needed it to be in different places for clarity.

javicarrillo (approved):

PR #302, opened 2021-11-25
Rework TypeaheadType

davidbeig:
The TypeaheadType is broken since the upgrade of Symfony. Now it is an extension of CollectionType with a base of TextType.


src/Goteo/Library/Forms/Admin/AdminPostEditForm.php
```diff
@@ -175,7 +155,7 @@ function($k, $v) {
                 'label' => 'admin-title-section',
                 'expanded' => true,
                 'wrap_class' => 'col-xs-6',
-                'choices' => $post::getListSections(),
+                'choices' => array_flip($post::getListSections()),
```
carlos-mg89, line 158: @davidbeig I've taken the chance to fix this form. Choices need a flip.

src/Goteo/Util/Form/Type/TypeaheadType.php
```diff
@@ -25,6 +25,8 @@ public function configureOptions(OptionsResolver $resolver)
     {
         parent::configureOptions($resolver);
         $resolver->setDefaults([
+            'entry_type' => TextType::class,
+            'allow_add' => true,
```
carlos-mg89, line 29: @davidbeig Looks like this was the only part needed to fix the TypeaheadType

src/Goteo/Model/Filter.php
```diff
@@ -741,7 +741,7 @@ public function getDonors($offset = 0, $limit = 0, $count = false, $lang = null)
                     $sqlFilter .= " AND YEAR(CURDATE())";
                 }
             } else if (isset($this->enddate)) {
-                $sqlFilter .= " AND donor.year <= :endyear ";
+                $sqlFilter .= " AND donor_cert.year <= :endyear ";
```
carlos-mg89, line 744: I've taken the chance to fix some issues with this method since it threw SQL errors. I would gladly show them to you.

src/Goteo/Util/Form/Type/TypeaheadType.php
```diff
@@ -25,13 +25,15 @@ public function configureOptions(OptionsResolver $resolver)
...
```
davidbeig: If we don't change the base type from TextType to CollectionType, maybe we don't need this two fields, right?

davidbeig (commented):
Great job! I still don't really get why it wasn't working without the MultipleTypeaheadType, but glad you managed to make it work :)

I just have the comment I made about the defaults of TypeaheadType, after it i think we can merge it!

carlos-mg89:
> Great job! I still don't really get why it wasn't working without the MultipleTypeaheadType, but glad you managed to make it work :)
> 
> I just have the comment I made about the defaults of TypeaheadType, after it i think we can merge it!

We need 2 different types, because in one case we need a string and in the other case we need a Collection (otherwise the array is not properly sent and validated).

src/Goteo/Util/Form/Type/TypeaheadType.php
carlos-mg89: Good spot! I've removed both of them here, and actually I was able to remove the `entry_type` from the `MultipleTypeaheadType` class too

carlos-mg89 (approved):

src/Goteo/Library/Forms/Model/FilterForm.php
```diff
@@ -57,8 +58,8 @@ public function createForm() {
             ->add('predefineddata', ChoiceType::class, [
```
carlos-mg89, line 58: @davidbeig what's this field doing? It's not on the SQL table (neither in past migrations) and it clearly doesn't get saved, even if I remove the `mapped` property. If it's not being used, I propose to remove it from the form.

carlos-mg89: Nevermind, I've just seen that it modifies the start and end dates.

javicarrillo (approved):

PR #308, opened 2021-12-08
Bug/fix api project endpoints

carlos-mg89:


src/Routes/api_routes.php
```diff
@@ -190,133 +176,122 @@
...
 $api->add('api-projects', new Route(
     '/projects',
-    array('_controller' => 'Goteo\Controller\Api\ProjectsApiController::projectsAction',
-        )
+    array('_controller' => 'Goteo\Controller\Api\ProjectsApiController::projectsAction',)
 ));
 
 // Projects by sdg
-$api->add('api-projects', new Route(
+$api->add('api-projects-by-sdg', new Route(
```
carlos-mg89, line 183: @davidbeig @javicarrillo the issue that broke the api projects search was here. An override of the 'api-projects' endpoint. With this change, the previous endpoint is up and working as before.

src/Routes/invest_routes.php
```diff
@@ -13,76 +13,61 @@
...
-    array('_controller' => 'Goteo\Controller\InvestController::loginAction',
-        )
+    array('_controller' => 'Goteo\Controller\InvestController::loginAction',)
 ));
-$routes->add('invest-signup', new Route(
+$invest->add('invest-signup', new Route(
     '/invest/{project_id}/signup',
-    array('_controller' => 'Goteo\Controller\InvestController::signupAction',
-        )
+    array('_controller' => 'Goteo\Controller\InvestController::signupAction',)
```
carlos-mg89, line 40: @javicarrillo could you confirm that this was an actual issue? I found out, after a tidy up of the routes files, that the `invest-login` and `invest-signup` endpoints didn't seems to be working at all. I've renamed the `$routes` variable into `$invest`, and now they should be working again.

carlos-mg89:
The only files that have real changes are `src/Routes/api_routes.php` and `src/Routes/invest_routes.php`.

The other ones, have formatting changes.

carlos-mg89:
I've just tested this PR in edge and it works like a charm.

src/Routes/invest_routes.php
davidbeig: Hi @carlos-mg89, it worked with the $routes. Since the file was being included directly inside the routes.php, the $routes variable was correct. By changing it to the $invest router, we have to change the route from '/invest/{project_id}/signup' to '/{project_id}/signup'. 

Thanks for pointing this up! I do it myself rn.

davidbeig (changes requested):

davidbeig (approved):

PR #315, opened 2021-12-13
GH Actions service change from mysql to mariadb

davidbeig:
Change gh actions service from mysql to mariadb

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,8 +19,8 @@ jobs:
         php: [ '7.4' ]
 
     services:
-      mysql:
-        image: mysql:8.0
+      mariadb:
+        image: mariadb:10.2
         ports:
           - 3306:3306
         env:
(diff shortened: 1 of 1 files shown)


microstudi (commented):
I think configuration for mariadb is a little bit different, something like this:

image

.github/workflows/test.yml
```diff
@@ -37,10 +37,11 @@ jobs:
           --health-retries=3
 
     steps:
-      - name: Verify MySQL connection from host
+      - name: Verify MariaDB connection from host
```
microstudi, line 40: an this step is probably all innecessary...

davidbeig:
Okay! I'll change it later. I was trying now to make it work on locally using [@nektos/act](https://github.com/nektos/act)

microstudi:
@davidbeig nice repo! didn't know it!

davidbeig:
Should update this PR with #316 or just close it.

What do you think? @microstudi @carlos-mg89

carlos-mg89:
> Should update this PR with #316 or just close it.
> 
> What do you think? @microstudi @carlos-mg89

I think that we should have either MySQL or MariaDB in all environments.

microstudi:
> I think that we should have either MySQL or MariaDB in all environments.

All MariaDB, we should change any mention to mysql in the docs as well

PR #324, opened 2022-01-04
Improvements in Impact Discover

davidbeig:
This PR has some improvements in the Impact Discover landing


javicarrillo (approved):

public/assets/sass/layouts/_impact_discover.scss
```diff
@@ -228,13 +228,18 @@
...
         a {
             cursor: pointer;
+            display: flex;
+            align-items: center;
+            max-height: 100%;
+            width: 100%;
 
             img.bg-project {
-                position:absolute;
+                // position:absolute;
```
carlos-mg89, line 237: I think it's better to remove lines than comment them. Otherwise, we end up with polluted code.

davidbeig: True! Fixed in #328

PR #326, opened 2022-01-10
Creation of DataSets inside Core

davidbeig:
We create the DataSet model following a new design pattern, rather the usual we have in Goteo.

This new models are not attached to the default Goteo Model, but they change the pattern. Now we create a DataSet model and a DataSetRepository to handle the crud operations of the model. This way we will have shorter models and we'll be able to have a better transition towards any ORM.

This PR also includes tests for the new Repository and the new DataSet model.


db/migrations/20220104165950_goteo_data_sets.php
```diff
@@ -0,0 +1,100 @@
...
+   * Return the SQL statements for the Up migration
+   *
+   * @return string The SQL string to execute for the Up migration.
+   */
+  public function getUpSQL()
+  {
+     return "CREATE TABLE `data_set` (
+            `id` BIGINT(20) UNSIGNED NOT NULL AUTO_INCREMENT,
+            `title` TEXT NULL,
+            `description` VARCHAR(10) NULL,
```
carlos-mg89: @davidbeig I don't know the details of this task, but 10 characters seem to be too little for a description column.

db/migrations/20220104165950_goteo_data_sets.php
```diff
@@ -0,0 +1,100 @@
...
+            `lang` VARCHAR(6) NOT NULL,
+            `url` VARCHAR(10) NOT NULL,
```
carlos-mg89: @davidbeig I don't know the details of this task, but 10 characters it's too short for a URL field. Only with the scheme "https://", there are already 8 characters.

db/migrations/20220104165950_goteo_data_sets.php
```diff
@@ -0,0 +1,100 @@
...
+            `image` VARCHAR(255) NULL,
+            `created` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+            `modified` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
```
carlos-mg89: The convention for created and modified fields, is to write them as "created_at" and "modified_at". Or "created_on" and "modified_on". 

More info:https://stackoverflow.com/questions/1379543/database-name-convention-datetime-column

src/Goteo/Model/DataSet.php
```diff
@@ -0,0 +1,123 @@
...
+use DateTime;
+use Goteo\Application\Config;
+
+class DataSet {
+
+    private int $id;
+    private ?string $title = null;
+    private ?string $description = null;
+    private string $lang;
+    private ?string $url = null;
```
carlos-mg89: In the DB table, url is NOT NULL.

src/Goteo/Model/DataSet.php
```diff
@@ -0,0 +1,123 @@
...
+    private ?Image $image = null;
+    private ?int $created = null;
+    private ?int $modified = null;
```
carlos-mg89: The "created" and "modified" columns are NOT NULL.

src/Goteo/Repository/DataSetRepository.php
```diff
@@ -0,0 +1,148 @@
...
+ namespace Goteo\Repository;
+
+ use Goteo\Application\Exception\ModelException;
+ use Goteo\Application\Exception\ModelNotFoundException;
+ use Goteo\Model\DataSet;
+ use PDO;
+ use PDOException;
+
+ class DataSetRepository extends BaseRepository {
+     protected ?string $Table = 'data_set';
```
carlos-mg89, line 20: If this property is not going to be used, we better remove it. The same in the `BaseRepository` class.

carlos-mg89: Property names should start with lowercase

src/Goteo/Repository/BaseRepository.php
```diff
@@ -0,0 +1,43 @@
...
+use Goteo\Application\App;
+use Goteo\Application\Config;
+use Goteo\Core\DB;
+use Goteo\Core\Exception;
+use Goteo\Library\Cacher;
+
+abstract class BaseRepository {
+    protected DB $db;
+    private Cacher $cacher;
+    protected ?string $Table = null;
```
carlos-mg89: Property names should start with lowercase

src/Goteo/Repository/BaseRepository.php
```diff
@@ -0,0 +1,43 @@
...
+    protected $model;
+
+    public function __construct() {
+		if ($cache_time = Config::get('db.cache.time')) {
+			$this->cacher = new Cacher('sql', $cache_time);
+		}
+
+		$this->db = new DB($this->cacher, App::debug() ? 2 : false);
+    }
```
carlos-mg89, line 24: This method code isn't properly indented.

src/Goteo/Repository/DataSetRepository.php
```diff
@@ -0,0 +1,148 @@
...
+             return $this->query($sql)->fetchAll();
+         }
+
+         $sql = "SELECT data_set.*
+                FROM data_set
+                LIMIT $limit
+                OFFSET $offset";
+
+        return $this->query($sql)->fetchAll(PDO::FETCH_CLASS, DataSet::class );
+    }
```
carlos-mg89, line 55: This method is actually running 2 different SQL queries. 

Let's have 2 different methods, so each method has only one query. One would be `getList()` and the other one `count()`.

This way, the method has one less parameter (`$count`), and the method name is more appropriate.

Actually, the `count()` method won't need the `LIMIT` nor the `OFFSET`. It'll simply be:

```
$sql = "SELECT count(data_set.*)
                FROM data_set";
```

src/Goteo/Repository/DataSetRepository.php
```diff
@@ -0,0 +1,148 @@
...
+
+         $dataSet = $this->query($sql, [$id])->fetchObject(DataSet::class);
+
+         if (!$dataSet instanceOf DataSet)
+             throw new ModelNotFoundException("DataSet with id $id not found");
+
+         return $dataSet;
+    }
+
+    public function getList(array $filter, int $offset = 0, int $limit = 10, bool $count = false) {
```
carlos-mg89: The `$filter` property is not being used, so we better remove it.

src/Goteo/Repository/DataSetRepository.php
```diff
@@ -0,0 +1,148 @@
...
+
+         $sql = "SELECT data_set.*
+                FROM data_set
+                LIMIT $limit
+                OFFSET $offset";
+
+        return $this->query($sql)->fetchAll(PDO::FETCH_CLASS, DataSet::class );
+    }
+
+    public function getListByFootprint(array $footprints) {
```
carlos-mg89: If `$footprints` is empty, I'm pretty confident that the SQL doesn't need even to be executed. So returning an empty list will most likely be more efficient. 

And in this case, adding the contents of `$sqlWhere` directly into the `$sql` will be more concise.

Or am I missing something @davidbeig ?

src/Goteo/Repository/DataSetRepository.php
```diff
@@ -0,0 +1,148 @@
...
+            $sqlWhere = "WHERE fds.footprint_id IN ( " . implode(',', $footprints) . ")";
+         }
+
+        $sql = "SELECT data_set.*
+                FROM data_set
+                INNER JOIN footprint_data_set fds ON fds.data_set_id = data_set.id
+                {$sqlWhere}
+                ";
+
+        return $this->query($sql)->fetchAll(PDO::FETCH_CLASS, DataSet::class );
```
carlos-mg89, line 74: @davidbeig are we returning an array of DataSet objects here? If so, we could add in the PHPDoc of this method, the following:

```
/**
* @return DataSet[]
**/
```

This way, the IDE will recognize that we are returning an array of DataSet, instead of a simple array.

src/Goteo/Repository/DataSetRepository.php
carlos-mg89: Shoudn't the method return an array?

src/Goteo/Repository/DataSetRepository.php
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+     public function getListBySDGs(array $sdgs) {
```
carlos-mg89: Same comments as int `getListByFootprint(array $footprints)`

src/Goteo/Repository/DataSetRepository.php
```diff
@@ -0,0 +1,148 @@
...
+         $sql = "SELECT data_set.*
+                FROM data_set
+                INNER JOIN sdg_data_set sds ON sds.data_set_id = data_set.id
+                {$sqlWhere}
+                ";
+
+         return $this->query($sql)->fetchAll(PDO::FETCH_CLASS, DataSet::class );
+     }
+
+     public function getListByCall(array $calls) {
```
carlos-mg89: Same comments as int `getListByFootprint(array $footprints)`

tests/Goteo/Model/DataSetTest.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+namespace  Goteo\Model\Tests;
+
+use Goteo\Model\DataSet;
+use Goteo\TestCase;
+
+class DataSetTest extends TestCase {
```
carlos-mg89: I don't think that we need to test that the constructor of `DataSet` returns a `Dataset`.

If we really want to test the `DataSet` class, we should test the setters and getters. That could be good to increase our project's code coverage.

src/Goteo/Repository/DataSetRepository.php
davidbeig: The $filter is not being used at the moment. I've defined it because it's how we used to create the getList functions before, as we could use this $filter array to include new conditions to the query, like searching by dates or whatever. We could change and then just create a getDataSetsBetweenDates or another function like that.

src/Goteo/Repository/DataSetRepository.php
davidbeig: Since the join is made with an INNER JOIN, executing the query will give us all the data sets related to footprints, so in this case it gives us some data that getList doesn't... it's true that right now we don't have a use case for this results...

tests/Goteo/Repository/DataSetRepositoryTest.php
```diff
@@ -0,0 +1,72 @@
...
+
+    /**
+     * @depends testGetList
+     */
+    public function testSaveDataSet(): DataSet {
+        $dataSet = new DataSet();
+        $dataSet->setId(1)
+            ->setTitle('Data Set 1.')
+            ->setDescription(' Description of DataSet 1')
+            ->setUrl(Config::get('url.main'));
```
carlos-mg89: Let's avoid calling static methods as much as possible. We can simply write "http://www.example.com" as an URL.

carlos-mg89: This way, we are removing one use statement (`Config` class) and reduce usages to static methods, that will be removed sooner than later.

tests/Goteo/Repository/DataSetRepositoryTest.php
```diff
@@ -0,0 +1,72 @@
...
+        $this->assertEquals($dataSet->getId(), $dbDataSet->getId());
+        return $dbDataSet;
+    }
+
+    /**
+     * @depends testDataSetExists
+     */
+    public function deleteDataSet(DataSet $dataSet): void {
+        $this->repository->delete($dataSet);
+        $this->assertCount(0, $this->repository->getList([]));
```
carlos-mg89: We could try following the GIVEN WHEN THEN structure.

In this case, the GIVEN comes from the parameter, but we can still have a WHEN block and a THEN block. Like this:

```
$this->repository->delete($dataSet);

$datasets = $this->repository->getList([]);
$this->assertCount(0, $datasets);
```

src/Goteo/Repository/DataSetRepository.php
carlos-mg89: I see what you mean, however, I still think that if is not needed right now, it should be removed. Adding it later is very simple and quick.

davidbeig:
After all this changes, we have decided that the new Model DataSets should be moved to a new Entity namespace, thus splitting up the new models to the older ones. @carlos-mg89

carlos-mg89 (dismissed):

src/Goteo/Entity/DataSet.php
```diff
@@ -0,0 +1,127 @@
...
+namespace Goteo\Entity;
+
+use DateTime;
+use Goteo\Application\Config;
+use Goteo\Model\Image;
+
+class DataSet {
+
+    const DATASET_TYPE_PROJECTS = 'projects';
+    const DATASET_TYPE_INVESTS = 'invests';
```
carlos-mg89: These 2 constants could be renamed to `TYPE_PROJECTS` and `TYPE_INVESTS`. Since they are in the `DataSet` entity class, we already have the context over there and repeating it in the name of the constant is redundant.

src/Goteo/Model/Traits/DataSetRelationsTrait.php
```diff
@@ -0,0 +1,100 @@
...
+        try {
+            self::query($sql, $values);
+        } catch (PDOException $e) {
+            throw new ModelException($e->getMessage());
+        }
+
+        return $this;
+    }
+
+    public function getAllDataSet($lang = null) {
```
carlos-mg89: Should return an array, and in the PHPDoc, an array of `DataSet`, right?

src/Goteo/Repository/BaseRepository.php
carlos-mg89: This method is still not rightly indented.

src/Goteo/Repository/DataSetRepository.php
```diff
@@ -0,0 +1,267 @@
...
+        try {
+            $this->query($sql, $values);
+        } catch (PDOException $exception) {
+            $errors[] = $exception->getMessage();
+        }
+        return $dataSet;
+    }
+
+    public function delete(DataSet $dataSet): void {
+        $sql = "DELETE FROM $this->table WHERE $this->table.id = :id";
```
carlos-mg89, line 260: Aren't we missing the single quotes, as used in the `update()` method above. For both `$this->table` and `$this->table.id`?

tests/Goteo/Repository/DataSetRepositoryTest.php
```diff
@@ -0,0 +1,89 @@
...
+        $this->assertEquals('Data Set 1 Update', $dataSet->getTitle());
+        return $dataSet;
+    }
+
+    /**
+     * @depends testDataSetExists
+     */
+    public function testDeleteDataSet(DataSet $dataSet): void
+    {
+        print_r("delete function");
```
carlos-mg89: Tests should run without printing anything in the console. Otherwise we end up with an output being difficult to read.

carlos-mg89 (approved):

PR #327, opened 2022-01-12
OpenData options for Sdgs and Footprints

davidbeig:
Create new inputs for the OpenData command to extract data related to Sdgs and Footprints projects and their invests.


src/Goteo/Console/Command/OpenDataCommand.php
```diff
@@ -223,6 +299,20 @@ private function extractCallInvestsData(Call $call): void {
...
+                $dataSetRepository->persist($dataSet);
+                $call->addDataSet($dataSet, 1);
+            } catch (ModelNotFoundException $e) {
+                $dataSet = new DataSet();
+                $dataSet->setTitle("$call->name");
+                $dataSet->setUrl($file->get_path() . $fileName);
+                $dataSet->setType(DataSet::DATASET_TYPE_INVESTS);
+                $dataSetRepository->persist($dataSet);
+                $call->addDataSet($dataSet, 1);
+            }
```
carlos-mg89, line 232: @davidbeig the `try` and `catch` blocks of these lines could be refactored to avoid code duplication. Specially the `catch` block, which is identical in all cases except for the argument passed to the `setType()` method.

We could have a method that creates the `DataSet`, with an `int $type` parameter, and reuse it later in all `catch` blocks of this `Command`.

The `try` block is still likely to have a refactor, with a little bit more of work, but definitely doable. This will save us a tone of lines of code and lots of work in a future refactor. :wink:

carlos-mg89:
@davidbeig tests are broken. There seems to be a missing Trait.

src/Goteo/Console/Command/OpenDataCommand.php
```diff
@@ -10,71 +10,320 @@
...
 
-            $call = Call::get($call_id);
+            try {
+                $call = Call::get($call_id);
+            } catch (ModelNotFoundException $e) {
+                $this->log("Call $call_id does not exist", [], 'error');
+                return;
+            }
+
             $this->extractCallOpenData($call);
```
carlos-mg89: This block would be nice to have it in a new method

src/Goteo/Console/Command/OpenDataCommand.php
```diff
@@ -10,71 +10,320 @@
...
         }
+
+        if ($listSdg = $input->getOption(('sdg'))) {
+            foreach ($listSdg as $sdg_id) {
+                $this->log("Retrieving {$sdg_id}'s data", [], 'info');
+
+                $sdg = Sdg::get($sdg_id);
+                $this->extractSdgOpenData($sdg);
+            }
```
carlos-mg89: This block would be nice to have it in a new method

src/Goteo/Console/Command/OpenDataCommand.php
```diff
@@ -10,71 +10,320 @@
...
+        }
+
+        if ($listFootprints = $input->getOption('footprint')) {
+            foreach ($listFootprints as $footprint_id) {
+                $this->log("Retrieving {$footprint_id}'s data", [], 'info');
+
+                $footprint = Footprint::get($footprint_id);
+                $this->extractFootprintOpenData($footprint);
+            }
```
carlos-mg89: This block would be nice to have it in a new method

src/Goteo/Console/Command/OpenDataCommand.php
```diff
@@ -10,71 +10,320 @@
...
+
+        $this->extractSdgProjects($sdg);
+        $this->extractSdgInvests($sdg);
+    }
+
+    private function extractSdgProjects(Sdg $sdg) {
+        $fileName = time() . '-' . $sdg->id . '-projects.csv';
+        $file = File::factory(['bucket' => AWS_S3_BUCKET_DOCUMENT]);
+        $file->connect();
+        $file->setPath("open_data/sdg/$sdg->id/projects");
```
carlos-mg89: These 4 lines are repeated 6 times in a pattern that we could have extracted as a method that returns a file. Example:

```
function getFile(int $id, string $type)
{
    $fileName = time() . "-$id-$type.csv";
    $file = File::factory(['bucket' => AWS_S3_BUCKET_DOCUMENT]);
    $file->connect();
    $file->setPath("open_data/sdg/$id/$type");
}
```

carlos-mg89: Reviewing the lines below, we might need another parameter, for the file path, since sometimes is not `open_data/sdg`, but `open_data/footprint`

carlos-mg89 (dismissed):
Looks great and tests are back to green ;)

Good job David!

carlos-mg89 (dismissed):
Looks great!

javicarrillo (approved):

PR #329, opened 2022-01-14
Implemented support for PeerTube videos on Project overview

carlos-mg89:


public/assets/js/forms.js
```diff
@@ -712,18 +678,21 @@ $(function(){
...
-                 .fail(function(e){
-                    // console.log('error vimeo', e.responseText);
-                 });
+                 .fail(function(e) { });
+            } else if (video.type === 'framatube') {
+                $.getJSON("https://peertube2.cpy.re/api/v1/videos/" + video.id)
+                    .done(function(res) {
+                        putVideo("https://framatube.org/lazy-static/previews/" + res.uuid + ".jpg");
+                    })
+                    .fail(function(e) { });
```
davidbeig: Could we change the use of this $.getJSON to a fetch call? I think that, even though we are not going to remove jquery any time soon, i don't think we should be using it in new features.

carlos-mg89:
I've migrated some code to a more recent alternative (JS fetch API, https://caniuse.com/fetch). It's supposed to be working with a big percentage of users. So I don't think it's an issue at all.

Same for JavaScript ES6 (https://caniuse.com/es6).

Both changes are supported by 94-95% of users worldwide. Seems reasonable to me that we move forward with these changes. Both changes are from 2015, I think we are definitely good to go in 2022 :)

@davidbeig @javicarrillo @microstudi do you agree to go this way? Compilation with `docker/exec grunt watch` works perfectly fine. And it's working fine for me in Firefox 96.

In Decidim, ES6 is enabled too: https://github.com/Platoniq/decidim/blob/40d47302d0dfb3506f998081894c7dbc43f43c41/packages/eslint-config/index.js#L4

public/assets/js/forms.js
carlos-mg89: Done!!! You can check it in the recent commit.

javicarrillo (approved):

PR #331, opened 2022-01-17
Value Module as Component

davidbeig:
This PR creates the Values Component. Previously it was created as a partial for the home. Since it is now required to be used in other layouts, like the one that uses Channel, we have decided to transform it into a component.

There is a new components/values as well as the necessary JS and CSS for the component in new files.

This PR also includes the use of the component in the Channel layout and the necessary migrations and new functions.


src/Goteo/Controller/ChannelController.php
```diff
@@ -134,6 +137,23 @@ public function indexAction($id)
...
+            foreach($footprints as $footprint) {
+                foreach ($footprint->getAllImpactData() as $impactData) {
+                    if ($channel->hasImpactData($impactData)) {
+                        $footprintImpactData[$footprint->id][] = $impactData;
+                    }
+                }
+                $projects_by_footprint[$footprint->id] = Project::getByFootprintOrSDGs(['footprints' => $footprint->id, 'channel' => $channel->id]);
+                $sdg_by_footprint[$footprint->id] = Sdg::getList(['footprint' => $footprint->id]);
+            }
+        }
```
carlos-mg89, line 155: It'll be nicer to have this inside a method. Perhaps, even in the `Project` and `Sdg` model. This way we keep the controller concise and brief.

There could be a method that obtains all Projects by `Footprint` in `Project` model. And another method that returns the `Sdg` by Footprint in the `Sdg` model.

src/Goteo/Model/Traits/ImpactDataRelationsTrait.php
```diff
@@ -48,6 +48,18 @@ public function addImpactData(ImpactData $impact_data, int $order) {
...
     }
 
+    public function hasImpactData(ImpactData $impactData): bool {
+        $tb = strtolower($this->getTable());
+        $rel = $this->getImpactDataTable();
+
+        $sql = "SELECT impact_data.*
+            FROM `$rel`
+            INNER JOIN impact_data ON impact_data.id = `$rel`.impact_data_id
+            WHERE `$rel`.impact_data_id = ?";
```
carlos-mg89, line 57: The SQL could be optimized by retrieving just the id in the SELECT. Since it could return a lot of information which is not relevant to know if there is related `ImpactData`.

Example:
```
SELECT impact_data.id
FROM `$rel`
INNER JOIN impact_data ON impact_data.id = `$rel`.impact_data_id
WHERE `$rel`.impact_data_id = ?
```

carlos-mg89:
@davidbeig I've left you a couple of comments that are very quick to address.

src/Goteo/Model/Traits/ImpactDataRelationsTrait.php
davidbeig: True! In fact, we don't even need the join with impact data, since we only want to check that it exists in the relation table.

src/Goteo/Controller/ChannelController.php
carlos-mg89: It'd be nicer to have this inside a method. Perhaps, even in the `Project` and `Sdg` model. This way we keep the controller concise and brief.

There could be a method that obtains all Projects by `Footprint` in `Project` model. And another method that returns the `Sdg` by Footprint in the `Sdg` model.

davidbeig: I have just extracted it from this concrete action and created a private method to add those variables as context to the engine.

javicarrillo (approved):

PR #337, opened 2022-01-25
Updated PeerTube preview to conform with latest URL changes for plata…

carlos-mg89:
…formess instance

### DIFF
--- a/public/assets/js/forms.js
+++ b/public/assets/js/forms.js
@@ -43,28 +43,31 @@ function parseVideoURL (url) {
 
     url.match(/(http:|https:|)\/\/(player.|www.|m.)?(vimeo\.com|youtu(be\.com|\.be|be\.googleapis\.com)|framatube\.org|peertube\.plataformess\.org)\/(videos\/watch\/|video\/|embed\/|watch\?v=|v\/|w\/)?([A-Za-z0-9._%-]*)(\&\S+)?/);
 
-    var type, src, thumbnailSrc = "";
+    let type, src, thumbnailSrc = "", apiUrl = "";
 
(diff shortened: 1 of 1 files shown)


public/assets/js/forms.js
```diff
@@ -659,21 +662,21 @@ $(function(){
...
-            var video = parseVideoURL(val);
-            var input = this;
-            var $container = $(this).closest('.media-container');
-            var $holder = $container.find('.video-holder');
-            var $embed = $container.find('.embed-responsive');
+            let video = parseVideoURL(val);
+            let input = this;
+            let $container = $(this).closest('.media-container');
+            let $holder = $container.find('.video-holder');
+            let $embed = $container.find('.embed-responsive');
```
davidbeig, line 671: Why create this fields as let instead of const? Following JS developers I see that they tend to use const in this cases instead of let. We could try to define a common ground for this JS developments. @carlos-mg89

davidbeig (commented):
Sorry for this comment after this PR being merged already. It's just to create a debate :P

public/assets/js/forms.js
carlos-mg89: I'd need to review deeper that block and see which fields should be dynamic (let) and which ones are not (const), but you're right @davidbeig 

To be honest, I simply updated the code from var to let without thinking very much about this. We should definitely sit down and talk about having a convention with this.

If you want, you could start a document to have the common JS (perhaps another for PHP too) and we could talk about it in the next Goteo meeting.

davidbeig: I sure will!

PR #354, opened 2022-02-16
Improve accesibility on images, links and navigation

davidbeig:
- add alt to linked images
- add alt to images inside the project layout
- add titles for inputs that had no title or label
- fixes on empty links
- fixes on links, images for updates on projects
- improve accesibility in project participation section

### DIFF
--- a/Resources/templates/responsive/auth/partials/form_signup.php
+++ b/Resources/templates/responsive/auth/partials/form_signup.php
@@ -5,28 +5,41 @@ $suggest = $this->suggest ? $this->suggest : [];
 ?>
 
 <div class="form-group<?= $err['name'] ? ' has-error' : '' ?>">
-    <input type="text" class="form-control" placeholder="<?= $this->text('register-name-field') ?>" name="name" value="<?= $this->name ?>" required>
+    <label for="name"><?= $this->t('register-name-field') ?></label>
+    <input id="name" type="text" class="form-control" name="name" value="<?= $this->name ?>" required>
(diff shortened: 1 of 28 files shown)


javicarrillo (approved):

Resources/templates/responsive/auth/partials/form_signup.php
```diff
@@ -5,36 +5,49 @@
...
+    <input id="name" type="text" class="form-control" name="name" value="<?= $this->name ?>" required>
     <?= ($err['name'] ? '<span class="help-block info-block">' . $err['name'] . '</span>' : '') ?>
 </div>
 
 <div class="form-group<?= $err['email'] ? ' has-error' : '' ?>">
-    <input type="email" class="form-control" placeholder="<?= $this->text('register-email-field') ?>" name="email" value="<?= $this->email ?>" required>
+    <label for="email">
+        <?= $this->text('register-email-field') ?>
+    </label>
+    <input id="email" type="email" class="form-control" name="email" value="<?= $this->email ?>" required>
```
carlos-mg89, line 17: @davidbeig  It's quite likely that we will get more errors related with having several inputs with the same ID (aside from the one you found, with the ID named "search").

In the future, I'll suggest that we name them using a prefix with the intent of the form.

Example:

If we are in form_signup, we could have:

- Instead of ID "email" => "user_signup_email"
- Instead of ID "name" => "user_signup_name"

The longer, the better. Let's not get crazy about this, but might be nice to bear this in mind when making this kind of changes in the future.

PR #368, opened 2022-03-08
Add DataSet component to channel layout

davidbeig:


javicarrillo:
There are conflicts to resolve @davidbeig.

Resources/templates/responsive/channel/partials/data_sets.php
```diff
@@ -0,0 +1,16 @@
+<?php
+if ($section = current($this->channel->getSections('data_sets'))):
+    $dataSets = $this->channel->getAllDataSet();
+?>
+
+<div class="section data-sets">
+    <div class="container">
+        <h2 class="title"><?= $section->main_title ?></h2>
```
javicarrillo: This title should be centered

javicarrillo (changes requested):

Resources/templates/responsive/channel/partials/data_sets.php
```diff
@@ -0,0 +1,14 @@
+<?php
+    $section = $this->dataSetsSection;
+?>
```
javicarrillo, line 4: I would add an "if" because the section could be active but empty the datasets channel

Resources/templates/responsive/channel/partials/data_sets.php
```diff
@@ -0,0 +1,14 @@
...
+
+<div class="section data-sets-section">
+    <div class="container">
+        <h2 class="title text-center"><?= $section->main_title ?></h2>
+        <p class="description">
+            <?= $section->main_description ?>
+        </p>
+        <?= $this->insert('partials/components/data_sets', ['dataSets' => $this->dataSets]) ?>
```
javicarrillo: I would change the data_sets component, in mobile the item should be 100% width.

src/Goteo/Model/Node/NodeSections.php
```diff
@@ -58,31 +60,26 @@ public static function getLangFields() {
 
 
   /**
-   * Get data about Node Sections
-   *
-   * @param   int    $id         check id.
-   * @return  NodeSections object
+   * @return  NodeSections
```
carlos-mg89: The return comment isn't necessary, since there's already a return parameter in this method

javicarrillo (approved):

PR #381, opened 2022-03-28
Symfony 4.0 upgrade - Move Translations to new default folder

davidbeig:
Move all the translation files from Resources/translations to a new translations default folder


crowdin.yml
```diff
@@ -1,3 +1,3 @@
 files:
-  - source: /Resources/translations/es/*.yml
-    translation: /Resources/translations/%two_letters_code%/%original_file_name%
```
carlos-mg89, line 3: Good catch @davidbeig ! I missed this file in my PR

src/Goteo/Application/Lang.php
```diff
@@ -112,14 +112,6 @@ static public function trans($id, array $parameters = array(), $locale = null) {
         return static::$translator->trans($id, $parameters, null, $locale);
     }
 
-    /**
-     * Handy method for the transChoice() function of the Symfony Component Translator
-     */
-    static public function transChoice($id, $number, array $parameters = array(), $domain = null, $locale = null) {
-        static::factory();
-        return static::$translator->transChoice($id, $number, $parameters, $domain, $locale);
-    }
```
carlos-mg89, line 121: I've already removed this method (it was supposed to be done in the Symfony 4.2 upgrade), since is not being used.

carlos-mg89 (approved):

PR #383, opened 2022-03-28
Deprecate Node Iframe

davidbeig:


src/Goteo/Model/Node.php
```diff
@@ -72,6 +71,11 @@ class Node extends \Goteo\Core\Model {
         $config
         ;
 
+    /**
+     * @deprecated
+     */
+    public ?string $iframe;
```
carlos-mg89, line 77: @javicarrillo @davidbeig If we don't need this property anymore, could we remove the related DB column perhaps and then the property?

PR #386, opened 2022-03-30
Allow users with no matched invest to match when donating

davidbeig:


javicarrillo (approved):

src/Goteo/Util/MatcherProcessor/DuplicateInvestMatcherProcessor.php
```diff
@@ -36,15 +33,15 @@ static public function getVarLabels() {
...
+    static public function getDesc(): string {
         return Text::get('matcher-duplicateinvest-rules');
     }
 
     /**
      * Checks if this invests has to be multiplied and
      * returns the amount to be added
      */
-    public function getAmount(&$error = '') {
+    public function getAmount(&$error = ''): int {
```
carlos-mg89, line 44: @davidbeig Could this amount be a float in some case? I have the feeling that we might return a rounded value here, from a float to int. And then we could have inconsistent / incomplete data somewhere else.

davidbeig: The invests done inside the platform are always integers. The minimum amount is 1€. Since the MatcherProcessors only allow for multiplications (x2, x3, etc) or the remaining amount will never be a float.

carlos-mg89: Wonderful then!

PR #390, opened 2022-04-06
Improve responsive design in Values component

davidbeig:


Resources/templates/responsive/partials/components/values.php
```diff
@@ -43,18 +43,24 @@
...
-                            <div class="">
+                            <div class="row">
                                 <?php if ($impact_data->image): ?>
-                                    <img
-                                        src="<?= $impact_data->getImage()->getLink(165,240,true) ?>"
-                                        alt="<?= $footprint->title ?>"
-                                        height="240"
-                                        width="165"
-                                        >
+                                    <div class="col-md-4 col-xs-4 col-xss-12">
```
carlos-mg89: @davidbeig I'm unable to see any `col-xss-` class at the Bootstrap page for version 3.x.

I've checked this page: https://getbootstrap.com/docs/3.4/css/#grid-options

Could it be that you actually meant `col-xxs-12` instead of `col-xss-12`?

It definitely looks like the `col-xxs-` prefix is available thanks to this NPM package that we are using across the whole project: https://www.npmjs.com/package/bootstrap-xxs

carlos-mg89 (approved):

PR #391, opened 2022-04-07
Migrate to Symfony 4.1

carlos-mg89:


davidbeig (approved):
Everything goes as expected! Symfony 4.4 here we go!

PR #393, opened 2022-04-08
Add related URL query string on Not Found exception so we can try to …

carlos-mg89:
…fix/remove broken URLs


davidbeig (approved):
Nice catch! This way we will be able to see what happens with all these errors and know exactly where the users are trying to enter :)

PR #388, opened 2022-04-04
Deprecate Paypal fields from Project and Contract

davidbeig:
The paypal fields in the Contract and in the Project were not being used. Now to activate donations through paypal the user has to activate the allowpp field inside the ProjectCampaingForm.

PR #396, opened 2022-04-08
Upgrade to Symfony 4.2

carlos-mg89:


src/Goteo/Library/Forms/Model/ProjectCampaignForm.php
```diff
@@ -25,23 +25,20 @@
...
+    public function getPhoneConstraints(): array
     {
-        $constraints = [];
-
-        if($field === 'paypal') {
-            $constraints[] = new Constraints\Email([
-                'checkMX' => true,
-                'checkHost' => true
-            ]);
-        }
```
carlos-mg89, line 37: Since this field is actually getting removed in another PR, I don't see the trouble of removing these constraint options (which are getting deprecated in this Symfony version).

For now, I'll simply remove them from this form, since it's mandatory to do so. In the other PR, the field will simply be removed from the form.

src/Goteo/Library/Forms/Model/ProjectCampaignForm.php
```diff
@@ -51,7 +48,6 @@ public function createForm() {
             ->add('one_round', ChoiceType::class, [
                 'disabled' => $this->getReadonly(),
                 'label' => 'costs-field-select-rounds',
-                'constraints' => $this->getConstraints('one_round'),
```
carlos-mg89, line 54: Is actually returning an empty array...

src/Goteo/Library/Forms/Model/ProjectCampaignForm.php
```diff
@@ -63,21 +59,20 @@ public function createForm() {
...
-                'constraints' => $this->getConstraints('paypal'),
+                'constraints' => $this->getPaypalConstraints(),
                 'disabled' => $this->getReadonly(),
                 'required' => false,
                 'attr' => ['help' => Text::get('tooltip-project-paypal')]
             ])
             ->add('spread', TextareaType::class, [
                 'label' => 'overview-field-spread',
                 'disabled' => $this->getReadonly(),
-                'constraints' => $this->getConstraints('spread'),
```
carlos-mg89, line 80: Is actually returning an empty array...

src/Goteo/Library/Forms/Model/ProjectCampaignForm.php
davidbeig: I think we talked about this previously, but I don't remember if we reached a consensus on wether we should use the stric mode or not. Let me add that to use the strict mode we have to add a `composer require egulias/email-validator`

`
$constraints[] = new Constraints\Email([
'mode' => 'strict'
});
`

If we don't want to go with this library, I think it's interesting to, at least, use the html5 option.

carlos-mg89: Alright, if we still wanna give support to the 'paypal' field to be saved, we should definitely keep using the `new Constraints\Email()`.

However, there's a PR to actually remove this field from the form, isn't it?

We could maybe accept this PR, letting the `new Constraints\Email()`, without the deprecated options (`checkMX` and `checkHost`, that require the Composer dependency as you've mentioned). And then, merge the PR that removes the field from the form. That way, we don't need to keep the checkMX and checkHost options. I agree with you, that leaving the `Constaints\Email` is important, as long as we keep the field in the form (and removed it thinking that we actually didn't care about the field).

Lets discuss this later today.

davidbeig: Yes, you are right. I was thinking more about other forms, not in this one particularly.

We can just let it with the `Constraints\Email` and then with #388 remove this field.

carlos-mg89: Cool! There are no other forms with the `checkMX` or `checkHost` options. So with this change we will be good to go. I'm going to commit a change with the constraint, and then we can safely publish this to production.

src/Goteo/Library/Forms/Model/ProjectCampaignForm.php
```diff
@@ -25,23 +25,20 @@
...
+            return [];
         }
+    }
 
-        return $constraints;
+    public function getPaypalConstraints(): array
+    {
+        return [
+            new Constraints\Email(),
+        ];
```
carlos-mg89, line 41: @davidbeig actually, I had already created a method that kept the `Constraints\Email` constraint.

davidbeig (approved):

PR #403, opened 2022-04-20
Create migrations and models for Faq Sections and Subsections

davidbeig:


db/migrations/20220413104717_goteo_new_faq.php
```diff
@@ -0,0 +1,121 @@
...
+      // add the faq-migration code here
+      // Create default slugs for everybody
+      foreach(Faq::query("SELECT id,title FROM faq")->fetchAll(\PDO::FETCH_OBJ) as $faq) {
+        $slug = Faq::idealiza($faq->title, false, false, 150);
+        try {
+            // If duplicate, let it null
+            Faq::query("UPDATE faq SET slug=:slug WHERE id=:id", [':id' => $faq->id, ':slug' => $slug]);
+        } catch(\PDOException $e) {
+            //
+        }
```
carlos-mg89, line 26: This UPDATE query looks important to me.

If by any chance, an exception is thrown, the query won't be executed and it might let the system in an inconsistent status (we might expect all FAQs to have a slug value).

Perhaps, we could write a log that indicates that there was an issue with this UPDATE query for a given FAQ id. And deal with it in order to fix it, and let every FAQ with a slug.

db/migrations/20220413104717_goteo_new_faq.php
```diff
@@ -0,0 +1,121 @@
...
+   * Return the SQL statements for the Up migration
+   *
+   * @return string The SQL string to execute for the Up migration.
+   */
+  public function getUpSQL()
+  {
+     return "
+
+      CREATE TABLE `faq_section` (
+            `id` BIGINT(20) UNSIGNED NOT NULL AUTO_INCREMENT,
```
carlos-mg89: Are we really going to have so many FAQ sections that we need a BIGINT for the ID?

Maybe it's enough with a simple INT(11) type.

The same applies to the other tables being created in this migration.

src/Goteo/Model/Faq.php
```diff
@@ -10,81 +10,171 @@
...
+
+        if(is_string($id)) {
+            $sql .= "WHERE faq.slug = :slug";
+            $values = [':slug' => $id];
+        } else {
+            $sql .= "WHERE faq.id = :id";
+            $values = [':id' => $id];
+        }
+
+         //die(\sqldbg($sql, $values));
```
carlos-mg89: We could remove the comment

src/Goteo/Model/Faq.php
```diff
@@ -10,81 +10,171 @@
...
         $title,
+        $subsection_id,
         $description,
         $order;
 
     static public function getLangFields() {
         return ['title', 'description'];
     }
 
+    // fallbacks to getbyid
```
carlos-mg89, line 35: Why is it necessary to have a fallback to the method `getById`?

src/Goteo/Model/Faq.php
```diff
@@ -10,81 +10,171 @@
...
+
         return $faq;
     }
 
     /*
-     * Lista de proyectos destacados
+     * Lista de faqs
      */
-    public static function getAll ($section = 'node', $lang = null) {
+    public static function getAll ($subsection = 'node', $lang = null) {
```
carlos-mg89: The `subsection_id` column, is actually a `BIGINT`.

`ALTER TABLE faq ADD COLUMN subsection_id BIGINT(20) UNSIGNED NOT NULL AFTER title;`

If we allow the `$subsection` parameter to be a string, or anything that's not an `INT`, it will always return an empty list. Or even worse, the query might crash / return an exception.

I propose to set the `$subsection` parameter, as an `INT`, and remove the default `node` value.

src/Goteo/Model/Faq.php
```diff
@@ -10,81 +10,171 @@
...
+                  faq.node as node,
+                  $fields,
+                  faq.subsection_id as subsection,
+                  faq.order
+              FROM faq
+              $joins
+              $sql
+              ORDER BY faq.order ASC
+              LIMIT $offset, $limit";
+        //die(\sqldbg($sql, $values));
```
carlos-mg89: This comment can be removed

src/Goteo/Model/Faq.php
```diff
@@ -10,81 +10,171 @@
...
+            $sql = "
+                SELECT
+                    count(faq.id)
+                FROM faq
+                $joins
+                $sql
+            ";
+            $query = static::query($sql, $values);
+            return $query->fetchColumn();
+        }
```
carlos-mg89: Could we remove the `$count` param from the `getList()` method, and create a `getListCount()` method, for example, so `getList()` is more concise?

src/Goteo/Model/Faq.php
```diff
@@ -10,81 +10,171 @@
...
+              LIMIT $offset, $limit";
+        //die(\sqldbg($sql, $values));
+        $query = static::query($sql, $values);
+        return $query->fetchAll(\PDO::FETCH_CLASS, __CLASS__);
+    }
+
     public function validate (&$errors = array()) {
         if (empty($this->node))
             $errors[] = 'Missing node';
             //Text::get('mandatory-faq-node');
```
carlos-mg89: Can we remove this line please?

src/Goteo/Model/Faq.php
```diff
@@ -10,81 +10,171 @@
...
 
-        if (empty($this->section))
-            $errors[] = 'Missing section';
+        if (empty($this->subsection_id))
+            $errors[] = 'Missing subsection';
             //Text::get('mandatory-faq-section');
```
carlos-mg89: Can we remove this line please?

src/Goteo/Model/Faq/FaqSection.php
```diff
@@ -0,0 +1,173 @@
...
+
+        if(is_string($id)) {
+            $sql .= "WHERE faq_section.slug = :slug";
+            $values = [':slug' => $id];
+        } else {
+            $sql .= "WHERE faq_section.id = :id";
+            $values = [':id' => $id];
+        }
+
+        //die(\sqldbg($sql, $values));
```
carlos-mg89: Can we remove this comment please?

src/Goteo/Model/Faq/FaqSection.php
```diff
@@ -0,0 +1,173 @@
...
+        $query = static::query($sql, $values);
+
+        $item = $query->fetchObject(__CLASS__);
+
+        if($item) {
+          return $item;
```
carlos-mg89: Indentation doesn't follow the 4 spaces used in the rest of the file

src/Goteo/Model/Faq/FaqSection.php
```diff
@@ -0,0 +1,173 @@
...
+        }
+
+        if ($count) {
+            $sql = "SELECT
+                  count(faq_section.id)
+              FROM faq_section
+              $joins
+              $sql";
+            $query = static::query($sql, $values);
+            //die(\sqldbg($sql, $values));
```
carlos-mg89: Can we remove this comment please?

src/Goteo/Model/Faq/FaqSection.php
```diff
@@ -0,0 +1,173 @@
...
+            return $query->fetchColumn();
+        }
```
carlos-mg89: Could we separate this in a different method? `getListCount()` for example.

src/Goteo/Model/Faq/FaqSection.php
```diff
@@ -0,0 +1,173 @@
...
+                  faq_section.slug as slug,
+                  faq_section.icon as icon,
+                  faq_section.banner_header as banner_header,
+                  faq_section.order
+              FROM faq_section
+              $joins
+              $sql
+              ORDER BY faq_section.order ASC
+              LIMIT $offset, $limit";
+         //die(\sqldbg($sql, $values));
```
carlos-mg89: Can we remove this comment please?

src/Goteo/Model/Faq/FaqSection.php
```diff
@@ -0,0 +1,173 @@
...
+        $query = static::query($sql, $values);
+        return $query->fetchAll(\PDO::FETCH_CLASS, __CLASS__);
+    }
+
+
+  public function getFaqbySection(int $limit = 0): array
+  {
+
+    return Faq::getList(['section'=>$this->id], 0, $limit);
+
```
carlos-mg89: Could we remove the empty lines?

src/Goteo/Model/Faq/FaqSection.php
```diff
@@ -0,0 +1,173 @@
...
+              $sql
+              ORDER BY faq_section.order ASC
+              LIMIT $offset, $limit";
+         //die(\sqldbg($sql, $values));
+        $query = static::query($sql, $values);
+        return $query->fetchAll(\PDO::FETCH_CLASS, __CLASS__);
+    }
+
+
+  public function getFaqbySection(int $limit = 0): array
```
carlos-mg89: Shouldn't the method name be `getFaqBySection`?

src/Goteo/Model/Faq/FaqSubsection.php
```diff
@@ -0,0 +1,149 @@
...
+        if ($count) {
+            $sql = "SELECT
+                          count(faq_subsection.id)
+                      FROM faq_subsection
+                      $joins
+                      $sql";
+
+            $query = static::query($sql, $values);
+            return $query->fetchColumn();
+        }
```
carlos-mg89, line 82: Could we separate these lines in a new method that only deals with returning the count?

src/Goteo/Model/Faq/FaqSubsection.php
```diff
@@ -0,0 +1,149 @@
...
+                  faq_subsection.id,
+                  faq_subsection.section_id,
+                  $fields,
+                  faq_subsection.order
+              FROM faq_subsection
+              $joins
+              $sql
+              ORDER BY faq_subsection.order ASC
+              LIMIT $offset, $limit";
+         //die(\sqldbg($sql, $values));
```
carlos-mg89, line 94: Can we remove this comment please?

src/Goteo/Model/Faq/FaqSubsection.php
```diff
@@ -0,0 +1,149 @@
...
+        $query = static::query($sql, $values);
+        return $query->fetchAll(\PDO::FETCH_CLASS, __CLASS__);
+    }
+
+  public function getFaqbySubsection($limit=0){
```
carlos-mg89: Should the name be `getFaqBySubsection`?

src/Goteo/Model/Faq/FaqSubsection.php
```diff
@@ -0,0 +1,149 @@
...
+
+    return Faq::getList(['subsection'=>$this->id], 0, $limit);
+
+  }
+
+  /**
+   * Save.
+   *
+   * @param   type array  $errors
+   * @return  type bool   true|false
```
carlos-mg89: Let's add array and bool as parameter type and return type instead. Won't you agree @davidbeig ?

src/Goteo/Model/Faq/FaqSubsection.php
```diff
@@ -0,0 +1,149 @@
...
+        $errors[] = "Faq subsection save error: " . $e->getMessage();
+        return false;
+    }
+  }
+
+    /**
+     * Validate.
+     *
+     * @param   type array  $errors
+     * @return  type bool   true|false
```
carlos-mg89: Let's add the parameter and return types instead of comments

tests/Goteo/Model/FaqTest.php
```diff
@@ -119,15 +149,32 @@ public function testDelete(Faq $ob): Faq
...
+    private static  function delete_faq_section(): bool
+    {
+        try {
+            $faqSection= FaqSection::getBySlug(self::$sectionData['slug']);
+            $faqSection->dbDelete();
+        }
+        catch(ModelNotFoundException $e) {
+        }
+        catch(\PDOException $e) {
+            error_log('PDOException on deleting test faq section! ' . $e->getMessage());
```
carlos-mg89: If there's an error while removing the FaqSection, the test should simply fail.

There's no need to pollute the PHPUnit logs. Right now the shows the message between the test execution and is not clear what's going on.

Tests shouldn't print anything.

carlos-mg89 (changes requested):
A few comments have been left. Requires more work before approval.

db/migrations/20220413104717_goteo_new_faq.php
davidbeig: @javicarrillo I think it was you who did this. I don't think the library we are using rolls back if the postup function does not work... But yes, at least we should log the errors.

db/migrations/20220413104717_goteo_new_faq.php
davidbeig: No, we won't. You are right

db/migrations/20220413104717_goteo_new_faq.php
javicarrillo: I think it isn't critical because we have in Faq model cover the case "slug empty". I would add log in a future, right now there isn't logs in migrations and it would be an other rework task.

src/Goteo/Model/Faq.php
davidbeig: I think it was meant to be used with the slug property, but since some faqs might not have it, we could ask directly with the id if slug is not set. Obviously we could ask first the getBySlug and if there is no faq then ask with the getById()...

src/Goteo/Model/Faq/FaqSection.php
davidbeig: Since it's a method inside FaqSection i think we can just call it getFaqs. Thanks for pointing it out.

src/Goteo/Model/Faq/FaqSection.php
```diff
@@ -59,23 +59,21 @@ static public function get($id): FaqSection
...
         $item = $query->fetchObject(__CLASS__);
 
-        if($item) {
-          return $item;
-        }
+        if(!$item)
+            throw new ModelNotFoundException("Faq section not found for ID [$id]");
 
-        throw new ModelNotFoundException("Faq section not found for ID [$id]");
+        return $item;
```
carlos-mg89: Great refactor @davidbeig !

src/Goteo/Model/Faq/FaqSection.php
```diff
@@ -111,63 +98,82 @@ static public function getList(array $filters = [], int $offset = 0, int $limit
...
+     {
+         if(!$lang) $lang = Lang::current();
+         list($fields, $joins) = self::getLangsSQLJoins($lang, Config::get('sql_lang'));
+
+         $filter = [];
+         $values = [];
+
+         if($filter) {
+             $sql = " WHERE " . implode(' AND ', $filter);
+         }
```
carlos-mg89, line 115: The variable `$filter` will always be empty (since it has just been defined as empty 2 lines above), so this IF clause is never going to happen.

The same happens in the `getList()` method. The `$filter` variable is always an empty array.

Either we should be checking for `$filters` (the method parameter), or this IF could just be removed.

I think that issue is that we should be checking if `$filters` has any item inside instead of `$filter`.

src/Goteo/Model/Faq/FaqSection.php
```diff
@@ -111,63 +98,82 @@ static public function getList(array $filters = [], int $offset = 0, int $limit
...
-
-  }
+    /**
+     * @return Faq[]
+     */
+    public function getFaqs(int $limit = 0): array
+    {
+        $count = Faq::getListCount(['section' => $this->id]);
+        return Faq::getList(['section'=>$this->id], 0, $count);
+    }
```
carlos-mg89: @davidbeig The parameter `$limit` is not being used inside the method. Maybe we could just remove it?

src/Goteo/Model/Faq/FaqSection.php
```diff
@@ -111,63 +98,82 @@ static public function getList(array $filters = [], int $offset = 0, int $limit
...
+            'id',
+            'name',
+            'slug',
+            'icon',
+            'banner_header',
+            'button_action',
+            'button_url',
+            'lang',
+            'order'
+            ];
```
carlos-mg89, line 159: @davidbeig I propose you to move the `$fields` variable of this method, into a class constant. With a name that highlights the purpose of properties being saved.

This way, the method is much smaller, and perhaps the constant can be reused somewhere else.

Something like:

```
const EDITABLE_PROPERTIES = [
            'id',
            'name',
            'slug',
            'icon',
            'banner_header',
            'button_action',
            'button_url',
            'lang',
            'order'
        ];
```

src/Goteo/Model/Faq/FaqSubsection.php
```diff
@@ -96,54 +96,68 @@ static public function getList(array $filters = [], int $offset = 0, int $limit
...
+        if (!$this->validate($errors))
+            return false;
+
+        $fields = [
+            'id',
+            'section_id',
+            'name',
+            'lang',
+            'order'
+            ];
```
carlos-mg89, line 143: @davidbeig maybe you would like to move this variable as a class constant too. As I proposed you in `FaqSection` class.

tests/Goteo/Model/FaqTest.php
```diff
@@ -161,20 +161,10 @@ static function tearDownAfterClass(): void {
...
-        catch(ModelNotFoundException $e) {
-        }
-        catch(\PDOException $e) {
-            error_log('PDOException on deleting test faq section! ' . $e->getMessage());
-            return false;
-        }
-
-        return true;
+        $faqSection = FaqSection::getBySlug(self::$sectionData['slug']);
+        $faqSection->dbDelete();
```
carlos-mg89, line 167: Perfect!

src/Goteo/Model/Faq/FaqSection.php
davidbeig: The `$filter` variable is to be filled by conditions if the $filters array has some data to be filtered with. Most `getList` functions inside our models work this way. It's true that we can just have this getList without filters and that it only returns data based on `$offset` and `$limit`

src/Goteo/Model/Faq/FaqSection.php
davidbeig: You are right, but for now i would just keep this way since it's the one done in most models. After all the necessary reworks we'll be switching to Doctrine and then we won't need this kind of functions.

davidbeig:
I have changed already all your requests.

carlos-mg89:
I still see some unresolved topics, however, if you think they aren't important, I'll gladly approve the PR :) But just in case, review the unresolved conversations and write me again please.

carlos-mg89 (approved):
It looks pretty good @davidbeig ! Good job!

davidbeig:
This PR is not meant to be merged. It will be superseeded by a new PR with all the migrations, admin and view of the new FAQS

PR #405, opened 2022-04-21
Added missing image asset

carlos-mg89:


davidbeig (commented):
I think that, in fact, the asset is not missing. We can find it inside the slick library. It seems that in their slick-theme.css the path used is not the right one.

I don't know if there would be another solution instead of adding the same file to our own assets. What do you think?

carlos-mg89:
Yes. Actually, this "fix" isn't working. But yeah, the file is there (cause I just get it from the slick vendor folder).

However, changing the library's CSS file, doesn't seem right to me. Maybe this is actually fixed in a more recent version of the library.

Will close this PR anyway.

davidbeig:
I think there is not a more recent version. We are using the 1.8 release and in their github page it says it's the latest...

PR #416, opened 2022-05-03
Create donor landing and handle of invest's origins

davidbeig:
- create donate landing action
- create invest origin model and repositories
- add styles for donate landing
- add fields to handle invest origin donations


db/migrations/20220429110112_goteo_invest_origin.php
```diff
@@ -0,0 +1,63 @@
...
+   * @return string The SQL string to execute for the Up migration.
+   */
+  public function getUpSQL()
+  {
+      /**
+       *     invest_origin
+      source
+      detail
+      allocated
+       */
```
carlos-mg89: We could clear this up

src/Goteo/Entity/Invest/InvestOrigin.php
```diff
@@ -0,0 +1,63 @@
...
+namespace Goteo\Entity\Invest;
+
+use Goteo\Model\Invest;
+
+class InvestOrigin
+{
+    private int $invest_id;
+    private string $source;
+    private string $detail;
+    private string $allocated;
```
carlos-mg89: `$allocated` could be null, as the SQL doesn't state NOT NULL. So this property should be `?string`. Unless the SQL should be NOT NULL, in that case, the model would be already fine.

src/Goteo/Entity/Invest/InvestOrigin.php
```diff
@@ -0,0 +1,63 @@
...
+        return $this->detail;
+    }
+
+    public function setDetail(string $detail): InvestOrigin
+    {
+        $this->detail = $detail;
+        return $this;
+    }
+
+    public function getAllocated(): string
```
carlos-mg89: If the property could be null, then this method should return ?string

src/Goteo/Entity/Invest/InvestOrigin.php
```diff
@@ -0,0 +1,63 @@
...
+    {
+        return $this->allocated;
+    }
+
+    public function setAllocated(string $allocated): InvestOrigin
```
carlos-mg89: If the property could be null, then this method's parameter should be of ?string type

src/Goteo/Repository/InvestOriginRepository.php
```diff
@@ -0,0 +1,153 @@
...
+    /**
+     * @return string[]
+     */
+    public function getSources(): array {
+        $sql = "SELECT invest_origin.source
+                FROM invest_origin
+                GROUP BY invest_origin.source";
+
+        return $this->query($sql)->fetchAll();
+    }
```
carlos-mg89: If source could have repeated values, maybe it'd be interesting to have a `SELECT DISTINCT(invest_origin.source)` instead.

src/Goteo/Repository/InvestOriginRepository.php
```diff
@@ -0,0 +1,153 @@
...
+    /**
+     * @return string[]
+     */
+    public function getDetails(): array {
+        $sql = "SELECT invest_origin.detail
+                FROM invest_origin
+                GROUP BY invest_origin.detail";
+
+        return $this->query($sql)->fetchAll();
+    }
```
carlos-mg89: If detail could have repeated values, maybe it'd be interesting to have a `SELECT DISTINCT(invest_origin.detail)` instead.

src/Goteo/Repository/InvestOriginRepository.php
```diff
@@ -0,0 +1,153 @@
...
+    /**
+     * @return string[]
+     */
+    public function getAllocateds(): array {
+        $sql = "SELECT invest_origin.allocated
+                FROM invest_origin
+                GROUP BY invest_origin.allocated";
+
+        return $this->query($sql)->fetchAll();
+    }
```
carlos-mg89: If allocated could have repeated values, maybe it'd be interesting to have a `SELECT DISTINCT(invest_origin.allocated)` instead.

tests/Goteo/Repository/InvestOriginRepositoryTest.php
```diff
@@ -0,0 +1,88 @@
...
+        $this->repository = new InvestOriginRepository();
+    }
+
+    public function testExceptionOnGetByInvestId(): void {
+        $id = 1;
+        $this->expectException(ModelNotFoundException::class);
+        $this->expectExceptionMessage("InvestOrigin with invest_id {$id} not found");
+
+        $this->repository->getByInvestId($id);
+
```
carlos-mg89, line 27: Could this line be removed please?

carlos-mg89 (changes requested):
Needs more thorough testing (tests fail). Although I was able to see the donate page and seems to work as expected.

Some minor cleaning could be done (comments left).

Overall, looks quite nice. I'm sure it can be finished quickly.

tests/Goteo/Repository/InvestOriginRepositoryTest.php
davidbeig: I think not... This test expects the ModelNotFound exception and the message written, so it has to execute the code that actually throws the esception.

As far as i understand, this is the correct way of doing this test.

carlos-mg89: I meant the empty line, line 27.

davidbeig: Ah ok! I though you meant the call to getByInvestId!

carlos-mg89 (dismissed):

carlos-mg89:
Now it looks great!

carlos-mg89 (approved):

PR #417, opened 2022-05-04
Fix missing copies in payment method selection

davidbeig:


carlos-mg89:
Looks perfect! Good catch of the missing usage of translation.

javicarrillo (approved):
Good work!

PR #422, opened 2022-05-06
Feature/workshop partner

javicarrillo:
Add sponsor type to workshops.

### DIFF
--- a/Resources/templates/responsive/workshop/index.php
+++ b/Resources/templates/responsive/workshop/index.php
@@ -9,8 +9,6 @@ $this->layout('workshop/layout', [
     'tw_image' => $meta_img
     ]);
 
-$footer_sponsors=$this->workshop->getSponsors("footer");
-
 $this->section('workshop-content');
 
 ?>
@@ -37,6 +35,6 @@ $this->section('workshop-content');
 
 <?= $this->workshop->event_type=='crowdcoop' ? $this->insert('workshop/partials/partner_singulars') : '' ?>
 
-<?= $footer_sponsors ? $this->insert('workshop/partials/partner_footer', ['footer_sponsors' => $footer_sponsors]) : '' ?>
+<?= $this->footer_sponsors ? $this->insert('workshop/partials/partner_footer', ['footer_sponsors' => $this->footer_sponsors]) : '' ?>
 
 <?php $this->replace() ?>
--- a/Resources/templates/responsive/workshop/index.php
+++ b/Resources/templates/responsive/workshop/index.php
@@ -9,6 +9,8 @@ $this->layout('workshop/layout', [
     'tw_image' => $meta_img
     ]);
 
+$footer_sponsors=$this->workshop->getSponsors("footer");
+
 $this->section('workshop-content');
 
 ?>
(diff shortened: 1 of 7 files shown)


Resources/templates/responsive/workshop/partials/partner_footer.php
```diff
@@ -0,0 +1,13 @@
+<div class="partner">
+	<div class="container">
+		<div class="row">
+			<?php foreach($this->footer_sponsors as $sponsor): ?>
+				<div class="col-xs-12">
```
davidbeig, line 5: Are the sponsors meant to be shown one for every col?

imatge

src/Goteo/Model/Workshop/WorkshopSponsor.php
```diff
@@ -21,6 +21,7 @@ class WorkshopSponsor extends \Goteo\Core\Model {
     $id,
     $workshop,
     $name,
+    $type,
```
davidbeig, line 27: Can we also add some const values for the different types? 

```php
const TYPE_FOOTER = 'footer';
const TYPE_SIDE = 'side';
```

src/Goteo/Model/Workshop.php
```diff
@@ -319,15 +319,15 @@ public function getPosts () {
         return $this->postsList;
     }
 
-    public function getSponsors() {
+    public function getSponsors($type="side") {
```
davidbeig: are we always going to filter by ```$type```?  Maybe we can use it as a filter and then add it in the ```$values``` only if it's present.

davidbeig (commented):
I think it looks good! I added some comments about the models and the template, after that i think we can approve it!

Resources/templates/responsive/workshop/partials/partner_footer.php
javicarrillo: Yes, the need really is for a single logo width 100%.
I have also preferred to anticipate that there will be several logos full width style maybe in the future..

src/Goteo/Model/Workshop.php
javicarrillo: Yes, this method is used only in views, all the sponsors have to be a type defined to be shown

src/Goteo/Model/Workshop/WorkshopSponsor.php
javicarrillo: Good idea, added.

Resources/templates/responsive/workshop/partials/partner_footer.php
```diff
@@ -0,0 +1,13 @@
...
+					<a target="_blank" href="<?= $sponsor->url ?>" class="center-block">
+	            		<img class="img-responsive no-title center-block" src="<?= $sponsor->getImage()->getLink(1200, 0, false) ?>" alt="<?= $sponsor->name ?>">
+	        		</a>
```
davidbeig, line 8: Here, for future developments, we could lazy load these images and also add them using a picture tag so that we can make the browser download only the image size that it's going to display.

Resources/templates/responsive/workshop/index.php
```diff
@@ -9,6 +9,8 @@
     'tw_image' => $meta_img
     ]);
 
+$footer_sponsors=$this->workshop->getSponsors("footer");
+
```
davidbeig: I forgot to ask if we could move this to the WorkshopController.php and just pass it to the view. After that we can publish it directly!

Everything works as expected :)

davidbeig (commented):
If you want to change the last comment about the Controller we can wait, if not we can approve it now.

Resources/templates/responsive/workshop/partials/partner_footer.php
javicarrillo: Yes, we can take into account lazy load in the future.

Resources/templates/responsive/workshop/index.php
javicarrillo: done! I was thinking the same

davidbeig (approved):

PR #425, opened 2022-05-09
Add new PeerTube domain instance

carlos-mg89:


davidbeig (approved):
We just have to figure out how to handle this new peertube instances in a scalable way...

Approved!

PR #428, opened 2022-05-16
Update NPM's lodash package

carlos-mg89:


davidbeig (commented):
Well done!

I have just checked that we are only using lodash in the Gruntfile.js file to do a merge between two arrays... maybe we can just change it to vanillya JS and get rid of this library?

carlos-mg89:
> Well done!
> 
> I have just checked that we are only using lodash in the Gruntfile.js file to do a merge between two arrays... maybe we can just change it to vanillya JS and get rid of this library?

@davidbeig We could use VanillaJS, yes. However, that will imply having to deal with duplicate elements, since merging 2 arrays with the concat method. And we definitely will have duplicate elements, isn't it?

You could read more about this here: https://stackoverflow.com/questions/1584370/how-to-merge-two-arrays-in-javascript-and-de-duplicate-items

The lodash merge function is pretty advanced and optimized. I think it's better to use help libraries like this one, that even deal with levels inside the array. The array that we're merging is a multidimensional one.

I totally agree with you that using VanillaJS is actually pretty nice these days. But sadly there are many things that imply reworks and adding other functions.

Lodash is a widely used library. You can check how many projects that we are using at the moment, that depend on lodash too.

davidbeig (approved):

PR #429, opened 2022-05-18
Fix anonym users can't be messaged when they didn't resign reward

carlos-mg89:


Resources/templates/responsive/dashboard/project/invests.php
```diff
@@ -239,7 +250,7 @@
             <strong>#{id}</strong> -
             <em title="{date}">{timeago}</em>
             <a href="{respond_url}" class="btn btn-sm btn-pink"><i class="fa fa-paper-plane-o"></i> <?= $this->text('dashboard-project-mailing-write_again') ?></a>
-            <span class="recipient">{recipients} <?= $this->text('dashboard-menu-projects-recipients') ?></recipient>
+            <span class="recipient">{recipients} <?= $this->text('dashboard-menu-projects-recipients') ?></span>
```
carlos-mg89, line 253: This was an HTML issue that I've take the chance to fix

Resources/templates/responsive/dashboard/project/invests.php
```diff
@@ -149,9 +154,17 @@
...
+                  && !($invest->resign && $invest->anonymous); ?>
+              <?php if ($canSendMessages) { ?>
+                  <a data-toggle="modal" href="#messageModal"
+                     data-user="<?= $invest->getUser()->id ?>"
+                     data-name="<?= $invest->getUser()->name ?>"
+                     class="send-private" title="<?= $this->text('support-send-private-message') ?>">
+                      <span><?= (int)$this->messages[$invest->getUser()->id] ?></span>
+                      <i class="icon-1x icon icon-partners"></i>
+                  </a>
+              <?php } ?>
```
carlos-mg89: This is where the issues is actually fixed

Resources/templates/responsive/dashboard/project/invests.php
```diff
@@ -132,7 +130,14 @@
...
           <td><?= date_formater($invest->invested) ?></td>
-          <td><?php if($uid): ?><img src="<?= $invest->getUser()->avatar->getLink(30, 30, true) ?>" alt="<?= $name ?>" class="img-circle"> <?= $name ?><?php else: ?><?= $this->text('regular-anonymous') ?><?php endif ?> </td>
+          <td>
+              <?php if ($uid) { ?>
+                  <img src="<?= $invest->getUser()->avatar->getLink(30, 30, true) ?>" alt="<?= $name ?>" class="img-circle">
+                  <?= $name ?>
+              <?php } else { ?>
+                  <?= $this->text('regular-anonymous') ?>
+              <?php } ?>
+          </td>
```
carlos-mg89: Refactor to improve reading, functionality doesn't change

Resources/templates/responsive/dashboard/project/invests.php
```diff
@@ -149,9 +154,17 @@
           <td><?= $address ?></td>
           <td><?= $extra_info ?></td>
           <td>
-            <?php if(!$this->project->userIsOwner($invest->getUser())&&!$invest->anonymous): ?>
-            <a data-toggle="modal" href="#messageModal" data-user="<?= $invest->getUser()->id ?>" data-name="<?= $invest->getUser()->name ?>" class="send-private" title="<?= $this->text('support-send-private-message') ?>"><span><?= (int)$this->messages[$invest->getUser()->id] ?></span> <i class="icon-1x icon icon-partners"></i></a>
-            <?php endif ?>
+              <?php $canSendMessages = !$this->project->userIsOwner($invest->getUser())
+                  && !($invest->resign && $invest->anonymous); ?>
+              <?php if ($canSendMessages) { ?>
```
davidbeig: could we write ifs with the correspondent : and endif forms? i think it's easier to understand if we are closing and if a foreach or whatever

davidbeig (changes requested):
Thanks for this fix! I just had one comment, but since you are coming back in two weeks i'll do it myself.

Resources/templates/responsive/dashboard/project/invests.php
carlos-mg89: We certainly could write the if that way.

It's a preference for me to use bracers instead of endif. In my opinion, the lines in between an IF opening and closing, should be maximum 4-5.

If that would be the case, it wouldn't really mind in what way we write IFs, FORs, etc.

Feel free to change it your self if you prefer it the other way.

PR #438, opened 2022-05-30
Add origin register in profile form

javicarrillo:
Add origin register in user profile


translations/en/profile.yml
```diff
@@ -22,6 +22,12 @@ profile-field-legal-entity-ngo: NGO
 profile-field-legal-entity-others: Other
 profile-field-legal-entity-person: 'Individual'
 profile-field-legal-entity-self-employed: Freelancer
+profile-field-origin-register: 'How did you meet Goteo?'
+profile-field-origin-register-workshop: 'In a training workshop'
+profile-field-origin-register-event: 'Public event'
+profile-field-origin-register-social-media: 'Social Media'
+profile-field-origin-register-previous-project: 'Una campaña en Goteo anterior'
```
carlos-mg89: This translation is in Spanish

translations/en/profile.yml
```diff
@@ -22,6 +22,12 @@ profile-field-legal-entity-ngo: NGO
...
+profile-field-origin-register-search: 'A previous Goteo campaign'
```
carlos-mg89: I think this translation should be "Using a search engine". And actually, looks like this translation is for `profile-field-origin-register-previous-project` translation.

src/Goteo/Library/Forms/Model/UserProfileForm.php
```diff
@@ -169,6 +181,13 @@ public function createForm() {
                 'required' => false,
                 'color' => 'cyan'
             ])
+            ->add('origin_register', ChoiceType::class, [
+                'label' => 'profile-field-origin-register',
+                'constraints' => $this->getConstraints('origin_register'),
```
davidbeig, line 185: It does not have any constraint, so we could remove this.

davidbeig (changes requested):
Well done! I just left a commend, since Carlos had already requested some changes.

At first i understood that this PR was for the origins of projects, not of users.

Also, are we planning to have some charts about the origins of users in the near future?

PR #443, opened 2022-06-03
Add missing SASS variable

carlos-mg89:


davidbeig (approved):
Thanks for fixing it! Good catch :)

PR #444, opened 2022-06-03
Refactor to fix call extras plugin

carlos-mg89:
This PR is required in order to allow proper functioning of this PR's call-extras plugin: https://github.com/GoteoFoundation/goteo-call-extras/pull/2

### DIFF
--- a/src/Goteo/Controller/Api/ChartsApiController.php
+++ b/src/Goteo/Controller/Api/ChartsApiController.php
@@ -73,7 +73,7 @@ class ChartsApiController extends AbstractApiController {
      * Simple projects info data specially formatted for D3 charts
      * COSTS
      */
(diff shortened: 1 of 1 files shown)


src/Goteo/Controller/Api/ChartsApiController.php
```diff
@@ -73,7 +73,7 @@ protected function getMatcher($matcher, $private = false) {
      * Simple projects info data specially formatted for D3 charts
      * COSTS
      */
-    public function projectCostsAction($id) {
+    public function projectCostsAction($id, Request $request) {
```
carlos-mg89, line 76: To let controllers that inherit from this controller use the `Request` parameter, the parent needs to have it too.

javicarrillo (approved):

PR #447, opened 2022-06-08
Update Composer dependencies (mostly minor or patch releases)

carlos-mg89:
This PR updates Composer dependencies. Mostly minor and patch releases. Apparently not any BC, or not at least in any of the most significant dependencies.

Reviewed latest minor releases of: Guzzle, PHPMailer, Omnipay/Common, PHPUnit, Symfony, geoip2/geoip2

PHPUnit tests continue passing.

It seems everything is working alright, including test payments on edge environment.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -8,16 +8,16 @@
     "packages": [
         {
             "name": "amphp/amp",
-            "version": "v2.6.0",
+            "version": "v2.6.2",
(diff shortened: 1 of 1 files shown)


davidbeig (approved):
Works good in my local environment and also in our edge environment. Thanks a lot!

PR #449, opened 2022-06-09
Experimentally upgraded Symfony 4.4 to 5.4

carlos-mg89:
Even though this branch seems to work pretty nicely on my local environment, after having removed the deprecated code, and PHPUnit tests continue in green, there's a major issue that is going to block the possibility to upgrade to Symfony 5.x (doesn't matter which branch of Symfony 5).

When we recently upgraded to Symfony 4.4, it was spotted that 2 important classes required attention on `src/Goteo/Util/Form/FormBuilder.php`. Basically, there was an important breaking code change in Symfony 5, that enforces all Symfony projects to use Twig as a templating engine.

More info here: https://symfony.com/blog/new-in-symfony-4-3-deprecated-the-templating-component-integration

What should happen in order to refactor the code and allow a Symfony 5.x upgrade?

Either a Twig refactor is accomplished, or we adapt the `TranslatorHelper` and `TemplatingExtension` classes, and potentially some more classes.

So far, I haven't been able to find any possible alternative to make the transition easier.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,25 +22,25 @@
         "iignatov/lightopenid": "~1.0",
         "paypal/adaptivepayments-sdk-php": "*",
         "paypal/merchant-sdk-php": "*",
-        "symfony/config": "4.4.*",
-        "symfony/console": "4.4.*",
-        "symfony/css-selector": "4.4.*",
-        "symfony/dependency-injection": "4.4.*",
-        "symfony/dom-crawler": "4.4.*",
-        "symfony/error-handler": "4.4.*",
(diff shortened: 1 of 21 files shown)


carlos-mg89:
I've attempted to include all needed PHP classes that were deprecated and removed in Symfony 5.

After doing that, there's an error while rendering the form (only form rendering is broken, other pages seem to work fine) which makes using forms completely unusable.

So it looks like tit's quite challenging to have a temporary support for Symfony 5 with some deprecated classes.

I'd discard this project in favor of a full Twig migration.

PR #453, opened 2022-06-13
Fix flow of invest when it fails and there is Invest Origin

davidbeig:


carlos-mg89 (dismissed):

javicarrillo:
It doesn't work for me, I've tested it with dummy method and I get a redirection to /donate .

Captura de pantalla de 2022-06-20 10-14-07

javicarrillo (changes requested):
It needs to be review, the redirection is not working.

davidbeig:
The code works, the only thing that it's meant to change the funcionality of the donation flow when there is a Invest origin, it does not change the behaviour for a normal donation.

javicarrillo:
Ok, I understand now, thanks by the fix in the main donation flow related with invest origin, that was the error really.

javicarrillo (approved):

PR #458, opened 2022-06-17
Add Tip in User Manage invest details

davidbeig:


javicarrillo (changes requested):
I would add the donate amount after the project title, and with the label tip donation, example:
€ 33 Mediterranean Fragility + € 2 tip donation

javicarrillo (approved):

PR #460, opened 2022-06-27
Add filter for projects in negotiation

davidbeig:
Add the possibility to filter project owners by checking that the project is in negotiation. With this development as it is, other filters ( donors for projects in negotiation, etc) might not make sense, since they won't have that project status filter developed... should be add them just in case?

### DIFF
--- a/src/Goteo/Model/Filter.php
+++ b/src/Goteo/Model/Filter.php
@@ -1380,6 +1380,11 @@ class Filter extends Model {
                 AND project.status = :project_status
                 ";
             $values[':project_status'] = $this->project_status;
+        } else if ($this->project_status == Project::STATUS_NEGOTIATION) {
+            $sqlFilter .= "
+                AND project.status = :project_status AND project.`id` NOT REGEXP '[0-9a-f]{32}'
+            ";
+            $values[':project_status'] = Project::STATUS_EDITING;
         }
 
         $this->projects = $this->getFilterProject($this->id);
@@ -1537,6 +1542,11 @@ class Filter extends Model {
                 AND project.status = :project_status
                 ";
             $values[':project_status'] = $this->project_status;
+        } else if ($this->project_status == Project::STATUS_NEGOTIATION) {
+            $sqlFilter .= "
+                AND project.status = :project_status AND project.`id` NOT REGEXP '[0-9a-f]{32}'
+            ";
+            $values[':project_status'] = Project::STATUS_EDITING;
         }
 
         $this->projects = $this->getFilterProject($this->id);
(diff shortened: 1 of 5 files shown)


src/Goteo/Model/Project.php
```diff
@@ -33,14 +33,15 @@ class Project extends Model {
     use Traits\SdgRelationsTrait;
 
     // PROJECT STATUS IDs
+    const STATUS_NEGOTIATION = -2;
```
javicarrillo: Improvement from review face to face:

 // STATUS TO BE USED TO FILTER PROJECT
    const STATUS_DRAFT       = -1; // is this really necessary?
    const STATUS_NEGOTIATION = -2; // filter status, project's real status is self::STATUS_EDITING

    // PROJECT STATUS IDs
    const STATUS_REJECTED    = 0;
    const STATUS_EDITING     = 1; // editing or negotiating (if draft)
    const STATUS_REVIEWING   = 2;
    const STATUS_IN_CAMPAIGN = 3;
    const STATUS_FUNDED      = 4;
    const STATUS_FULFILLED   = 5;
    const STATUS_UNFUNDED    = 6;

javicarrillo (approved):

src/Goteo/Model/Project.php
```diff
@@ -32,15 +32,18 @@ class Project extends Model {
 
     use Traits\SdgRelationsTrait;
 
+    // STATUS TO BE USED TO FILTER PROJECT
+    const STATUS_DRAFT = -1; // is this really necessary?
+    const STATUS_NEGOTIATION = -2; // filter status, project's real status is self::STATUS_EDITING
```
carlos-mg89, line 37: Just a simple comment. I find it very awkward that there are status with a negative sign.

IMHO, there should be a change in the status, making them all start from 0. And then perform an SQL update query that modifies the lines accordingly.

src/Goteo/Model/Filter.php
```diff
@@ -1380,6 +1380,11 @@ public function getPromoters($offset = 0, $limit = 0, $count = false, $lang = nu
                 AND project.status = :project_status
                 ";
             $values[':project_status'] = $this->project_status;
+        } else if ($this->project_status == Project::STATUS_NEGOTIATION) {
+            $sqlFilter .= "
+                AND project.status = :project_status AND project.`id` NOT REGEXP '[0-9a-f]{32}'
+            ";
+            $values[':project_status'] = Project::STATUS_EDITING;
```
carlos-mg89, line 1387: These lines are duplicate code of below lines (or the other way around) 1545-1549.

There would be a reduced code maintenance if they would be in a common method. At least the `$sqlFilter` one.

PR #461, opened 2022-06-27
Create Unify Certificates Command

davidbeig:
- create new certificate to superseed old ones
- add loop to update certs of all users


javicarrillo (dismissed):
I think is ok, I have tested in my computer and it works.

javicarrillo (approved):

PR #463, opened 2022-06-29
Enable pure Symfony controllers

carlos-mg89:
A new base controller has been added that extends a Symfony base controller class (`AbstractController`).

This new `BaseSymfonyController`, allows rendering Foil and Twig templates.

A refactor has been applied so there are 2 TemplateEngines (they implement Symfony's `EngineInterface`) and that way the project continues getting closer to the standard Symfony components.

If you try to use the `TwigEngine`, it already loads a base Twig template that is under `/templates` folder.

When this PR gets approved, I'd propose you to refactor the maximum amount of controllers to inherit from the new `BaseSymfonyController`. It might require some changes, but it's a start.


src/Routes/misc_routes.php
```diff
@@ -29,6 +29,11 @@
         )
 ));
 
+$misc->add('about-test', new Route(
+    '/about/test',
+    ['_controller' => 'Goteo\Controller\TestController::testAction']
+));
```
carlos-mg89: This route mapping should be removed before accepting the PR. It's just to ease the testing of it.

At the moment, I wasn't able to load controller actions using Symfony Route annotations. Perhaps in another task...

src/Goteo/Controller/TestController.php
```diff
@@ -0,0 +1,24 @@
...
+*
+* For the full copyright and license information, please view the README.md
+* and LICENSE files that was distributed with this source code.
+*/
+
+namespace Goteo\Controller;
+
+use Symfony\Component\HttpFoundation\Response;
+
+class TestController extends BaseSymfonyController {
```
carlos-mg89: This controller should be removed before accepting the PR. It's just to ease the testing of it.

src/Goteo/Controller/BaseSymfonyController.php
```diff
@@ -0,0 +1,99 @@
...
+
+    public function dbCache($cache = null): bool
+    {
+        return DB::cache($cache);
+    }
+
+    public function dbReplica($replica = null): bool
+    {
+        return DB::replica($replica);
+    }
```
carlos-mg89, line 98: This should be removed when the project gets ridden of the current DB class

src/Routes/misc_routes.php
```diff
@@ -29,6 +29,16 @@
...
 
+$misc->add('about-test-foil', new Route(
+    '/about/test-foil',
+    ['_controller' => 'Goteo\Controller\TestController::testFoilAction']
+));
+
+$misc->add('about-test-twig', new Route(
+    '/about/test-twig',
+    ['_controller' => 'Goteo\Controller\TestController::testTwigAction']
+));
```
carlos-mg89: To be removed before merging PR

src/routes.php
```diff
@@ -25,6 +25,10 @@
     '/',
     array('_controller' => 'Goteo\Controller\IndexController::indexAction')
 ));
+$routes->add('new_home', new Route(
+    '/',
+    array('_controller' => 'Goteo\Controller\IndexController::indexAction')
+));
```
davidbeig, line 31: Why duplicate this route?

davidbeig (changes requested):
Good job @carlos-mg89 !

I have tested the new endpoints created and they work correctly. I think we can move on to delete those endpoints and plan new developments with this new controller

src/routes.php
carlos-mg89: That's something that should be fixed in a future. The 'home' route is somehow not available to the `Router` / `UrlGenerator`.

When I added this new_home endpoint, it worked. So I guess there's some issue with either naming a route 'home' or with the route path being '/'.

If you prefer so, we could simply remove the newly created 'new_home'. Just be aware that 'home' is not working.

davidbeig (approved):
Thanks for the changes! I'm approving it now :)

PR #468, opened 2022-07-06
Improvements in templates

davidbeig:
Change in the behavior of the console watcher to send templates on different days. Also it creates new templates and the .yml for each one of them, but only in spanish.

### DIFF
--- a/src/Goteo/Console/EventListener/ConsoleWatcherListener.php
+++ b/src/Goteo/Console/EventListener/ConsoleWatcherListener.php
@@ -307,7 +307,7 @@ class ConsoleWatcherListener extends AbstractListener {
                 break;
 
             // Extra
-            case 26: // Send information about contract in order to prepare documentacion
+            case 27: // Send information about contract in order to prepare documentacion
                 $this->info("Sending information about contract", [$project, 'days_active' => $days_active, 'days_funded' => $days_funded]);
                 $this->send($project, "14_days", ['owner']);
                 break;
--- a/src/Goteo/Console/EventListener/ConsoleWatcherListener.php
+++ b/src/Goteo/Console/EventListener/ConsoleWatcherListener.php
@@ -204,8 +204,12 @@ class ConsoleWatcherListener extends AbstractListener {
             case 3: // Una acción a diario, por pequeña que sea
             case 4: // Llama a todas las puertas
             case 5: // Busca dónde está tu comunidad
-            case 8: // Agradece en público e individualmente
                 $this->send($project, "tip_$days_active", ['owner']);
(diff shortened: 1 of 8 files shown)


javicarrillo (changes requested):
There are two yaml files empty.

javicarrillo (dismissed):

src/Goteo/Console/EventListener/ConsoleWatcherListener.php
```diff
@@ -228,11 +229,7 @@ public function onProjectActive(FilterProjectEvent $event) {
             // comprobación periódica pero solo un envío
             case 7: // Apóyate en quienes te van apoyando, si más de 20 cofinanciadores
                 // o en cuanto llegue a 20 backers (en fechas libres)
-            case 14:
-            case 17:
-            case 21:
-            case 24:
-            case 27:
+            case 26:
```
carlos-mg89, line 232: Case 26 is at line 300 on the new file. Previously, it was only once.

src/Goteo/Console/EventListener/ConsoleWatcherListener.php
```diff
@@ -316,59 +308,25 @@ function ($a, $b) {
 
             // Extra
             case 26: // Send information about contract in order to prepare documentacion
```
carlos-mg89: This case 26 will never be reached, since there's already one at line 232

src/Goteo/Console/EventListener/ConsoleWatcherListener.php
```diff
@@ -204,18 +204,19 @@ public function onProjectActive(FilterProjectEvent $event) {
             case 3: // Una acción a diario, por pequeña que sea
```
carlos-mg89, line 204: I really think we should be using the constants that are in the Template class instead of using the hardcoded values of those constants.

davidbeig: You are right, but since this values refer to days instead of values with other meanings... we could still create a DAY_X constant for every day.

carlos-mg89: Alright, I really thought they were template IDs. And not days. Nevermind about this comment then.

src/Goteo/Console/EventListener/ConsoleWatcherListener.php
davidbeig: True. I think that the day 27 is free, do you think we can move it one day? @javicarrillo

javicarrillo (approved):

PR #512, opened 2022-11-23
Create Impact Calculator landing

davidbeig:


javicarrillo (changes requested):
If you could push the commits with the new route would be great to testing, thanks!

src/Routes/project_routes.php
```diff
@@ -50,6 +50,11 @@
     array('_controller' => 'Goteo\Controller\ProjectController::createAction')
 ));
 
+$prjs->add('project-impact-calculator', new Route(
+    '/impact_calculator',
```
javicarrillo: Change to middle -

javicarrillo (approved):

PR #522, opened 2023-02-27
New logo Platoniq Foundation

javicarrillo:
Include new Platoniq Foundation logo


davidbeig:
Shouldn't we change the references of the previous logo to the new one as well?

PR #524, opened 2023-03-06
Fix/docker commands for mac

### LINKED ISSUE #523: Docker compatibility for MacOS
UID is a reserved variable for MacOS and it already includes the user id you get with `id -u`

When you try to export this variable on MacOS you receive this error: `docker/up: line 2: UID: readonly variable`

fblupi:
Fixes #523

### DIFF
--- a/docker/exec
+++ b/docker/exec
@@ -1,5 +1,3 @@
 #!/bin/bash
-if [[ "$OSTYPE" =~ ^linux ]]; then
-	export UID=`id -u`
-fi
+export UID=`id -u`
 docker-compose exec -u goteo php "$@"
--- a/docker/exec
+++ b/docker/exec
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 if [[ "$OSTYPE" =~ ^linux ]]; then
 	export UID=`id -u`
 fi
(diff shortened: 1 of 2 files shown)


docker/exec
```diff
@@ -1,3 +1,5 @@
 #!/bin/sh
```
davidbeig:
```suggestion
#!/bin/bash
```

docker/exec
```diff
@@ -1,3 +1,5 @@
...
-export UID=`id -u`
+if [[ "$OSTYPE" =~ ^linux ]]; then
+	export UID=`id -u`
+fi
```
davidbeig: I think that with the previous change we won't be needing this.

Turns out that UID is a variable used in bash, since this script was declared to be executed with /bin/sh it needed to be declared. One problem we are facing right now is that, in linux, /bin/sh is a link to dash, and before it was a link to /bin/bash.

Since dash does not support "[[ condition ]]" this script fails executing in ubuntu 22.04, maybe in your macOs /bin/sh is a link to /bin/bash and then it's not failing.

I suggest we change the declaration and remove this lines, what do you think?

docker/up
```diff
@@ -1,5 +1,7 @@
 #!/bin/sh
-export UID=`id -u`
+if [[ "$OSTYPE" =~ ^linux ]]; then
+	export UID=`id -u`
+fi
```
davidbeig: The same as in the other file.

davidbeig (changes requested):

maukoese:
Any idea on when this will be resolved?

PR #528, opened 2023-03-10
Disable enter click on impact data project form

davidbeig:


javicarrillo (dismissed):
It's ok for me, I've tested it in local. It could be improved adding enter to move between inputs, but better in other fix.

PR #530, opened 2023-03-17
Add platoniq foundation references

davidbeig:


javicarrillo (commented):
We must use Platoniq Foundation logo, not Platoniq association logo, you can use directly the oauth foundation logo.

javicarrillo (approved):

PR #532, opened 2023-03-24
Feature/permanent campaigns

davidbeig:


javicarrillo (changes requested):
I have tried to test it, and it doesn't work, the code is broken. 
You have no defined the method isPermanent() in the project model.

src/Goteo/Console/Command/RoundCommand.php
```diff
@@ -104,6 +103,12 @@ protected function execute(InputInterface $input, OutputInterface $output) {
         $action_done = false;
         $return_code = 0;
         foreach ($projects as $project) {
+
+            if ($project->type != Project\Conf::TYPE_CAMPAIGN) {
+                $this->warning("This project's type is not campaign. There are no actions to be taking regarding rounds");
```
javicarrillo: I would add some text to the message warning "..is not a campaign, it is a continuous fundraising project".

Resources/templates/responsive/project/widgets/partials/data_list_permanent.php
javicarrillo, line 1: We need to review this view, I would remove it from this PR

javicarrillo (changes requested):

PR #539, opened 2023-04-14
Improve Pages

davidbeig:
- create new page admin
- add new functions to Page model
- move about controller to responsive


javicarrillo (changes requested):
Fixes:

- Routes /press / workshop error
- Admin error, it needs switch between html and markdown editor

PR #549, opened 2023-07-06
Fix styles for side-widget in project partials

davidbeig:
The social rewards list has lost it's styles after we implemented some improvements in accessibility.

imatge

This PR tries to fix this applying the styles to the new html tags chosen for the list of social rewards. Previously the list was a set of divs, and every item was just another set of divs with the inner information. Currently the list is created with the ul and li tags, and inside every item we have an article and a header for the title.

imatge

### DIFF
--- a/Resources/templates/responsive/project/partials/side.php
+++ b/Resources/templates/responsive/project/partials/side.php
@@ -69,20 +69,21 @@
     </ul>
 
     <ul class="collaborations list-unstyled">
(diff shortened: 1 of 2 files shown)


rxnetwalker (approved):

rxnetwalker (approved):

rxnetwalker (approved):
Great!

PR #563, opened 2023-10-19
Channel Blog

davidbeig:
This PR allows the channel to have it's own blog.

This way when a channel shows the post that have a relation with the channel, it can show them under the `/channel/{channel_id}` route.

This change will also allow to, when having a domain redirected to a channel, keep the blog under the same domain.
This way we can have an `example.org` domain that redirects to `/channel/example`, and then have `example.org/channel/example/blog/{post_slug}`


Resources/templates/responsive/channel/call/layout.php
```diff
@@ -14,13 +14,14 @@
...
 
 <?php
-
 $this->append();
 
 $this->section('header');
+?>
 
-//Header empty
+<?php $this->insert('channel/call/partials/banner_header'); ?>
```
subiabre: Why is the php closing tag in line 20 if we reopen php in line 22 with no non-php code in between? Same for the php closing tag in this line and the next.

Resources/templates/responsive/channel/call/partials/banner_header.php
```diff
@@ -24,14 +25,14 @@
 			<div class="row">
 				<div class="col-lg-6 col-md-7 col-sm-8 col-xs-12">
 					<div>
-						<span class="title" style="<?= $this->colors['header'] ? "color:".$this->colors['header'].";" : '' ?> <?= $this->colors['header_shadow'] ? "text-shadow:none" : '' ?>" >
-						<?= $this->channel->subtitle ?>
+						<span class="title" style="<?= $colors['header'] ? "color:".$colors['header'].";" : '' ?> <?= $colors['header_shadow'] ? "text-shadow:none" : '' ?>" >
```
subiabre: Can we use single quotes (') instead of double quotes for the HTML values in here? E.g `<?= $colors['header'] ? 'color:'.$color['header'].';'`.

It would help readability and very slightly improve performance since PHP doesn't need to evaluate those strings.

Resources/templates/responsive/channel/call/partials/navbar.php
```diff
@@ -1,15 +1,18 @@
...
+    $colors = $this->a($this->colors);
+?>
 <div class="custom-header">
 	<div class="pull-left">
 		<a href="<?= '/channel/'.$this->channel->id . $this->lang_url_query($this->lang_current())?> ">
 			<img src="<?= $this->channel->logo ? $this->channel->logo->getlink(0,40) : '' ?>" height="40px">
 		</a>
 	</div>
 	<div class="pull-right hidden-xs">
-		<span style="<?= $this->colors['header'] ? "color:".$this->colors['header'] : '' ?> ">
```
subiabre, line 8: Single quotes for the raw css in html.

subiabre (changes requested):

subiabre (dismissed):

subiabre (approved):

PR #565, opened 2023-10-20
Add channel favicon

davidbeig:
Allow channels to have their own favicon based on their label field.

We create a new favicon section inside the layout of the application so that any child can add or rewrite the favicon of the page.

### DIFF
--- a/Resources/templates/responsive/channel/call/layout.php
+++ b/Resources/templates/responsive/channel/call/layout.php
@@ -23,6 +23,13 @@ $this->section('header');
 <?php
 $this->replace();
 
+$this->section('favicon');
+
+if ($this->channel->label): ?>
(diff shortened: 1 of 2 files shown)


Resources/templates/responsive/channel/call/layout.php
```diff
@@ -23,6 +23,13 @@
 <?php
 $this->replace();
 
+$this->section('favicon');
+
+if ($this->channel->label): ?>
+        <link rel="icon" type="image/x-icon" href=<?= $this->channel->label->getLink(16,16,true) ?>
```
subiabre: Missing closing `>` for the `link` tag. Maybe also quote the href value for better readability.

```php
<link ... href="<?= $this->channel->label->getLink(16, 16, true) ?>">
```

subiabre (changes requested):

subiabre (approved):

PR #568, opened 2023-10-27
Stripe Integration

subiabre:
This PR adds the following behaviour:

* New payment method: Stripe. Subscription support only. `StripeSubscriptionPaymentMethod` implements the integration. It calls the Omnipay gateway implemented for stripe subscriptions at `Omnipay\Stripe\Subscription` which handles the actual logic to create subscriptions at Stripe.
* Projects marked as permanent campaign that have enabled Stripe as payment method can now allow to receive donations via Stripe subscriptions.
* Rewards flagged as subscribable items to allow Project owners to specify monthly amounts.


db/migrations/20231010140732_goteo_subscriptions.php
```diff
@@ -0,0 +1,53 @@
...
+    /**
+     * Return the SQL statements for the Up migration
+     *
+     * @return string The SQL string to execute for the Up migration.
+     */
+    public function getUpSQL()
+    {
+        return "
+            ALTER TABLE `reward` ADD COLUMN `subscribable` int(1) DEFAULT NULL;
+            ALTER TABLE `project_account` ADD COLUMN `allow_stripe` int(1) DEFAULT NULL AFTER `allowpp`;
```
davidbeig: Can we make them default to false?

src/Goteo/Controller/StripeController.php
```diff
@@ -0,0 +1,65 @@
...
+        switch ($event->type) {
+            case Event::TYPE_INVOICE_PAID:
+                $this->createInvest(json_decode($request->getContent()['data']['object']));
+                break;
+            case Event::TYPE_INVOICE_PAYMENT_FAILED:
+                break;
+            case Event::TYPE_CUSTOMER_SUBSCRIPTION_DELETED:
+                break;
+            default:
+                break;
```
davidbeig: Why using a switch in this case? Are we going to use the other cases or are we going to leave them blank? Maybe we can just use an `if`

src/Goteo/Payment/Method/StripeSubscriptionPaymentMethod.php
```diff
@@ -0,0 +1,153 @@
...
+
+    public function refund(): ResponseInterface
+    {
+        throw new PaymentException("Refund not yet supported for subscriptions", 1);
+    }
+
+    public function calculateCommission($total_invests, $total_amount, $returned_invests = 0, $returned_amount = 0)
+    {
+        // Confirm Stripe commission
+    }
```
davidbeig: We can't merge until we know about this comissions, right?

src/Omnipay/Stripe/Message/SubscriptionRequest.php
```diff
@@ -0,0 +1,95 @@
...
+use Stripe\StripeClient;
+
+class SubscriptionRequest extends AbstractRequest
+{
+    private array $data;
+    private StripeClient $stripe;
+
+    public function __construct(array $data)
+    {
+        $this->data = $data;
```
davidbeig: If i'm not wrong, this refer to a User instance and a Invest instance, do you prefer to keep it as a private array or have two private fields one for each Model?

src/Goteo/Application/Config.php
```diff
@@ -255,6 +255,8 @@ static private function setDirConfiguration() {
 		Payment::addMethod('Goteo\Payment\Method\PaypalPaymentMethod');
 		// Adding Cash non-public payment method (manual admin investions)
 		Payment::addMethod('Goteo\Payment\Method\CashPaymentMethod', true);
+		// Adding Stripe payment method
+		Payment::addMethod('Goteo\Payment\Method\StripeSubscriptionPaymentMethod');
```
davidbeig: This change requires that the `settings.yml` has the key `stripe`  inside the `payments` section. We should update also the documentation to make sure that it is by default in every application. If not, we can change it and only add it if the stripe payment is active (But it will change completely the current configuration of the payments)

src/Routes/dashboard_routes.php
```diff
@@ -338,6 +338,11 @@
     ['_controller' => 'Goteo\Controller\Dashboard\ImpactProjectDashboardController::removeImpactItemProjectCostAction']
 ));
 
+$dash->add('dashboard-project-subscription', new Route(
+    '/project/{pid}/subscription',
+    ['_controller' => 'Goteo\Controller\Dashboard\ProjectDashboardController::subscriptionAction']
+));
+
```
davidbeig: I think we don't need this one rn.

src/Goteo/Controller/Dashboard/ProjectDashboardController.php
```diff
@@ -90,6 +91,11 @@ static function createProjectSidebar(Project $project, $zone = '', &$form = null
                 ['text' => '<i class="fa fa-2x fa-sliders"></i> 6. ' . Text::get('project-campaign'), 'link' => $prefix . '/campaign', 'id' => 'campaign', 'class' => $validation->campaign == 100 ? 'ok' : 'ko'],
                 ['text' => '<i class="icon icon-2x icon-supports"></i> ' . Text::get('dashboard-menu-projects-supports'), 'link' => $prefix . '/supports', 'id' => 'supports'],
             ];
+
+            if ($project->isPermanent()) {
+                $steps[] = ['text' => '<i class="fa fa-2x fa-money"></i>' . Text::get('dashboard-menu-projects-subscription'), 'link' => $prefix . '/subscription', 'id' => 'subscription'];
+            }
+
```
davidbeig: Not needed.

davidbeig (changes requested):
After this changes we can try to publish it in beta and test it there.

Thanks for the effort!

src/Goteo/Controller/StripeController.php
subiabre: Yes, we need to handle at least `TYPE_INVOICE_PAID` (subscription active) to grant access to rewards, `TYPE_INVOICE_PAYMENT_FAILED` and `TYPE_CUSTOMER_SUBSCRIPTION_DELETED` (user didn't pay or cancelled subscription) to reject access. So far blank until subscriptions get tied to rewards.

src/Omnipay/Stripe/Message/SubscriptionRequest.php
subiabre: It'd be ideal to have two private fields for each Model, unfortunately the Omnipay library trusts us to use this `$data` field to perform further operations under the hood.

src/Omnipay/Stripe/Subscription/Message/SubscriptionRequest.php
```diff
@@ -0,0 +1,95 @@
...
+    public function sendData($data)
+    {
+        $price = $this->stripe->prices->create([
+            'unit_amount' => ($data['invest']->amount + $data['invest']->donate_amount) * 100,
+            'currency' => 'eur',
+            'recurring' => ['interval' => 'month'],
+            'product' => $this->getStripeProduct($data['invest'])->id
+        ]);
+
+        $subscription = $this->stripe->subscriptions->create([
```
subiabre: TODO: Update the configuration passed to new subscriptions to automatically charge the customer after payment method is set.

src/Omnipay/Stripe/Subscription/Message/SubscriptionRequest.php
```diff
@@ -0,0 +1,111 @@
...
+        $price = $this->stripe->prices->create([
+            'unit_amount' => ($data['invest']->amount + $data['invest']->donate_amount) * 100,
+            'currency' => 'eur',
+            'recurring' => ['interval' => 'month'],
+            'product' => $this->getStripeProduct($data['invest'])->id
+        ]);
+
+        $session = $this->stripe->checkout->sessions->create([
+            'customer' => $this->getStripeCustomer($data['user'])->id,
+            'success_url' => $this->getRedirectUrl('/dashboard/subscriptions'),
```
davidbeig: When the invest is successfull, it should be redirected to the step 3 of the invest process, where the user can update their fiscal information and so on.

src/Omnipay/Stripe/Subscription/Message/SubscriptionRequest.php
```diff
@@ -0,0 +1,158 @@
...
+    /**
+     * Get the ID of the reward for the invest, or a given string in case of no reward
+     * @param Invest $invest
+     * @param string $noReward The string to return in case of no reward selected
+     * @return string
+     */
+    private function getInvestReward(Invest $invest, string $noReward): string
+    {
+        return !empty($invest->getRewards())
+            ? $invest->getRewards()[0]
```
davidbeig: This line is broken. It returns an object. Since the signature of the function expects a string, this should return

`$invest->getFirstReward()->id`

davidbeig (changes requested):
I think we only need to this fix and then we can publish it.

src/Omnipay/Stripe/Subscription/Message/SubscriptionRequest.php
subiabre: Done!

davidbeig (approved):

PR #571, opened 2023-10-31
Feature/project post

davidbeig:
#### :tophat: What? Why?
Allow project owners to handle the privacy settings for project post based on the User donations to the project rewards.

### :camera: Screenshots

:hearts: Thank you!


src/Goteo/Library/PostPrivacyAccess.php
```diff
@@ -0,0 +1,42 @@
...
+        $admin = Session::isAdmin();
+        if ($admin)
+            return true;
+
+        if ($user instanceof User) {
+            if ($user->id == $post->author)
+                return true;
+
+            if ($post->access_limited) {
+                $postRewardAccessList = PostRewardAccess::getList(['post' => $post->id]);
```
subiabre: Shouldn't the key for the filters array be `post_id` as defined in https://github.com/GoteoFoundation/goteo/blob/feature/project_post/src/Goteo/Model/Blog/Post/PostRewardAccess.php#L85

src/Goteo/Model/Invest.php
```diff
@@ -2466,4 +2466,17 @@ public function getInvestOrigin(): ?InvestOrigin
             return null;
         }
     }
+
+    static public function hasInvestToReward(User $user, Reward $reward): bool
```
subiabre: This code would grant access to subscription-based rewards regardless of the subscription status. For those rewards we need to check if there's an invest that is up to date with the subscription.

db/migrations/20231030152314_goteo_create_post_reward_access.php
```diff
@@ -0,0 +1,61 @@
...
+   *
+   * @return string The SQL string to execute for the Up migration.
+   */
+  public function getUpSQL()
+  {
+      // return sql query to create table that links post with reward to grant access to the post content
+     return "
+        ALTER TABLE `post` ADD COLUMN `access_limited` TINYINT(1) UNSIGNED NOT NULL DEFAULT 0 AFTER `allow`;
+
+        CREATE TABLE `post_reward_access` (
```
subiabre, line 38: Please add a DROP statement for the new column and table before attempting to create them to ensure the migration runs properly.

Resources/templates/responsive/dashboard/project/privacy.php
```diff
@@ -0,0 +1,47 @@
...
+                </tr>
+            <?php endforeach ?>
+              </tbody>
+            </table>
+
+    <?= $this->insert('partials/utils/paginator', ['total' => $this->total, 'limit' => $this->limit]) ?>
+
+    <?php else: ?>
+        <blockquote>
+            <?= $this->text('dashboard-project-blog-privacy-empty') ?>
```
subiabre, line 41: Don't forget to add the text for this key ❗❗

Resources/templates/responsive/project/updates.php
```diff
@@ -47,33 +47,75 @@
...
+                    <a class="pronto"
+                       data-pronto-target="#project-tabs"
+                       href="<?= '/project/'.$this->project->id.'/updates/'.$update->post->getSlug() ?>">
+                        <?= $update->post->title ?>
+                    </a>
+                </h2>
+
+                <strong>
+                    <i class="fa fa-2x fa-lock"></i>
+                    You can't access this resource. To be able to do so you have to donate to some of these rewards:
```
subiabre: Please use the translation

src/Goteo/Library/PostPrivacyAccess.php
```diff
@@ -0,0 +1,60 @@
...
+        if ($user->id == $post->author)
+            return true;
+
+        if ($post->access_limited) {
+            $postRewardAccessList = PostRewardAccess::getList(['post_id' => $post->id]);
+            foreach($postRewardAccessList as $postRewardAccess) {
+                $reward = $postRewardAccess->getReward();
+                if ($reward->subscribable) {
+                    $hasAccessToPost = Invest::hasUserInvestToRewardInTheLastMonth($user, $reward);
+                    if ($hasAccessToPost)
```
subiabre: Deciding to return `true` only when `$hasAccessToPost` is `true` makes the function continue to the next `else` statement, making it possible at this part of the code to obtain an erroneous `true` value when trying to determine if a user has access to a subscription-based reward when it skips to the method `hasUserInvestToReward` which returns `true` on any amount of successful invests regardless of the time window.

I think this functionality can be accomplished much more elegantly by:
1. Refactor `hasUserInvestToReward` to hold the logic to check if the given `$reward` is a subscription-based one, then do the necessary lookup for that case inside the method.
2. Simply return the value of the method `hasUserInvestToReward`, e.g:
```php
$reward = $postRewardAccess->getReward();
return Invest::hasUserInvestToReward($reward);
```

subiabre (changes requested):

db/migrations/20231030152314_goteo_create_post_reward_access.php
davidbeig: What do you mean? They should not exists prior to execution. None of the migration have a DROP before any getUpSql() function.

Do you mean adding them to the preUp() function?

Since the migration can't be run selectively, but they run based on the date i don't know how this could be a problem.

src/Goteo/Library/PostPrivacyAccess.php
davidbeig: I don't get the part about the falsy true, I don't get how it will jump to the next `else`, since the `if` is about the subcribability of the reward :S

About the refactoring, why should the Invest class know that if the reward is subscribable it should check only for Invests in the previous month, or if there are invests at all?

I think this is part of the knowledge of the PostPrivacyAccess "library".

I think that we could work on a new function to extract this info directly from the database with a sql query, instead of generating so many database requests. If you want we can give it a try now, or leave it for the future.

subiabre: My bad, as I was reading the if I thought it would skip due to a syntax misreading. Please do refactor it to avoid `if ($hasAccessToPost)` and simply return the value of the privacy check method.

Regarding on where to put the logic, I'm OK with the two methods we discussed, for reference:

1. Create an InvestRepository to hold the SQL to search for invests to specific rewards.
2. Use an approach similar to Symfony's Voters: https://symfony.com/doc/current/security/voters.html

subiabre (approved):

subiabre (dismissed):

PR #573, opened 2023-11-06
Feature archive rewards

davidbeig:
#### :tophat: What? Why?
Allow admins to archive rewards. This means that the rewards can't be seen in the project layout and in the invest layout.

This feature is needed by some projects in order to hide exhausted rewards or focus on newer rewards.

#### Testing
Enter the rewards module of a project and, as admin, check or uncheck the archived checkbox. See that the reward can't be seen in the two views mentioned previously.

### :camera: Screenshots
*Please add screenshots of the changes you are proposing*
Reward with the archived checkbox active

:hearts: Thank you!


Resources/templates/responsive/invest/partials/reward_box_resign.php
```diff
@@ -1,5 +1,3 @@
-    <form class="form-horizontal" role="form" method="GET" action="/invest/<?= $this->project->id ?>/payment">
```
subiabre, line 1: What is the purpose of removing the `<form>` tag from this partial to be declared on insertion? Specially when this partial only holds form markup.

Also please, note the orphaned closing `</form>` tag left at https://github.com/GoteoFoundation/goteo/blob/d19276f610b4c3d76098a4f44df89099cedcd98c/Resources/templates/responsive/invest/partials/reward_box_resign.php#L23

If going ahead with decoupling `<form>` declaration from partial, I think it's best to also have to close the form markup after partial insertion for better code readability.

Resources/templates/responsive/invest/select_reward.php
```diff
@@ -14,25 +14,21 @@
...
-                    <?php //check if show the exhausted rewards ?>
-                    <?php if($reward_item->available()||!$this->project::hideExhaustedRewards($this->project->id)||!$this->project->inCampaign()): ?>
-
-                        <?= $this->insert('invest/partials/reward_box', ['reward_item' => $reward_item]) ?>
-
-                    <?php endif; ?>
-
-                <?php endforeach ?>
+                    <?php foreach($this->rewards as $reward_item):
+                        if ($reward_item->archived) continue;
```
subiabre: In my opinion, archived rewards are also unavailable rewards. Can't we update the `$reward_item->available()` method to also check if the reward is archived? That way we can also remove this double check and further ones in the code.

subiabre (changes requested):

Resources/templates/responsive/invest/select_reward.php
davidbeig: The contract for the `available` function is related to the relation between the reward units available from the project owner and the number of times the reward has been taken by a donation. Therefore i don't think we can change it for another functionallity. We can still unify it with the following if statement.

subiabre (approved):

PR #579, opened 2023-11-17
Add 'creator' role

subiabre:
#### :tophat: What? Why?
This PR adds the role 'creator' with permissions to highlight a project in their creator profile.

### DIFF
--- a/Resources/roles.yml
+++ b/Resources/roles.yml
@@ -12,6 +12,12 @@ user: # This user is assigned automatically to all users regardless the user_rol
         - remove-project # remove own project, this permission is implicit
     level: 0 # Order of the role, roles with supperior level can do some actions over this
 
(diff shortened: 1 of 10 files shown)


davidbeig (changes requested):
Something is not working as it should be. When I add the "content creator" to an admin profile, then the admin module will not show all the modules (all the content for an admin).

imatge

subiabre:
> Something is not working as it should be. When I add the "content creator" to an admin profile, then the admin module will not show all the modules (all the content for an admin).
> 
> imatge

This behavior is not exclusive to the "creator" role. The same thing happens with other roles such as "vip". Can you confirm this is in fact unintended behavior for every role?

davidbeig (approved):

PR #587, opened 2023-12-14
Redirect to a second checkout on subscriptions with platform tip

subiabre:
#### :tophat: What? Why?
This PR fixes the previously undefined behaviour of tips on subscriptions payments. See https://app.asana.com/0/1206149597772083/1206163960878458

#### Testing
If you choose to do a donation to the platform when paying with a subscription via Stripe, you'll be redirected to a second checkout (for the donation) after the first checkout (for the subscription).

### DIFF
--- a/src/Omnipay/Stripe/Subscription/Message/SubscriptionRequest.php
+++ b/src/Omnipay/Stripe/Subscription/Message/SubscriptionRequest.php
@@ -5,8 +5,10 @@ namespace Omnipay\Stripe\Subscription\Message;
 use Goteo\Application\Config;
 use Goteo\Library\Text;
 use Goteo\Model\Invest;
+use Goteo\Model\Project;
 use Goteo\Model\User;
 use Omnipay\Common\Message\AbstractRequest;
+use Stripe\Checkout\Session as CheckoutSession;
 use Stripe\Customer;
 use Stripe\Product;
 use Stripe\StripeClient;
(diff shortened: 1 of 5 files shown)


src/Omnipay/Stripe/Subscription/Message/SubscriptionRequest.php
```diff
@@ -34,40 +37,35 @@ public function sendData($data)
...
+            'customer' => $customer,
+            'success_url' => $successUrl,
+            'cancel_url' => $this->getRedirectUrl('project', $project->id),
+            'mode' => CheckoutSession::MODE_SUBSCRIPTION,
             'line_items' => [
                 [
-                    'price' => $price->id,
+                    'price' => $this->stripe->prices->create([
+                        'unit_amount' => $invest->amount * 100,
+                        'currency' => 'eur',
```
davidbeig: Should we change this for the settings field Currency?

subiabre: Updated this checkout to use the project's currency and the platform tip checkout to use Config currency. This way each checkout matches the payments recipient's (project and platform) currency.

davidbeig (changes requested):
When doing a new invest i found a bug. After paying the first donation, the one that represents the subscription, it will create a new Invest entry. At the end of the process we end with two invests, one with the amount + donation_amount, and another with only the amount, both with valid status.

davidbeig:
I belive the problem comes from the `StripeSubscriptionController.php`

After both invest creation, the controller does not look for a previously created invest based on the metadata of the invest. We should look for it and then update it. When the first donation is correct then we should update the donation_amount to 0, and then when the second payment is correct then return the value.

What do you think? @subiabre

rxnetwalker (approved):

PR #592, opened 2023-12-18
Update invests that have been refunded by Stripe

subiabre:
#### :tophat: What? Why?
Invests generated via Stripe subscriptions weren't being updated on the platform after they were refunded on Stripe. Now they are.

#### :pushpin: Related Issues
Fixes https://app.asana.com/0/1206149597772083/1206194873937420

#### Testing
Make a payment using the subscription with Stripe payment method.

Forward Stripe events to your local webhook using the CLI.

Then refund this payment and check your investment on /admin. Status should be cancelled.

:hearts: Thank you!


davidbeig (approved):
Works great!

PR #594, opened 2024-01-16
Allow to filter backer invests by subscription based payment methods

subiabre:
#### :tophat: What? Why?
Add a new filter on the "Backers" project dashboard section to filter invest records by subscription based payment methods.

#### :pushpin: Related Issues
https://app.asana.com/0/1205628877590802/1206101369621065

:hearts: Thank you!


src/Goteo/Model/Invest.php
```diff
@@ -492,6 +493,13 @@ public static function getSQLFilter($filters = []) {
                     $sqlFilter[] = "invest.resign = 0";
                     $sqlFilter[] = "invest.campaign = 0";
                     break;
+                case 'from_subscription':
+                    $values[':subscription_methods'] = implode(',', [
+                        StripeSubscriptionPaymentMethod::PAYMENT_METHOD_ID
+                    ]);
+
+                    $sqlFilter[] = "invest.method IN (:subscription_methods)";
```
davidbeig: Are we expecting to have more Subscription methods? Since we currently only have one, and the filter is in singular, i would prefer that we check that the value is the filtered value. Afterwards we can add a new filter to handle multiple subsriptions methods.

Also why can't we use the already created 'method' filter? It uses an array as input, so it also works like the one you coded.

We can still approve this PR since it's working correctly.

subiabre: We are not expecting any more subscription methods in a concrete sense, but it's known that the operational cost of Stripe is high and we are currently researching other payment gateways that also have subscription capabilities, so while we are not "oficially" expecting any more subscription methods, I wouldn't be surprised if the need to implement a new one pops up in the future.

On the other, yeah, I hadn't noticed the `method` filter could be used for this same purpose. Seems better to me to have one less filter. I'm repurposing the code to use the `method` filter.

davidbeig (approved):

PR #595, opened 2024-01-16
Fix participate invests from subscriptions

subiabre:
#### :tophat: What? Why?
Displays different text on invests listed in the participate section if they come from subscription based payment methods.

#### :pushpin: Related Issues
https://app.asana.com/0/1206149597772083/1206194536007200/f

### :camera: Screenshots
imagen

:hearts: Thank you!

### DIFF
--- a/Resources/templates/responsive/project/participate.php
+++ b/Resources/templates/responsive/project/participate.php
@@ -1,8 +1,4 @@
-<?php
-
-use Goteo\Payment\Payment;
-
- $this->layout('project/layout') ?>
+<?php $this->layout('project/layout') ?>
 <?php $this->section('main-content') ?>
 
 <?php $project=$this->project ?>
@@ -57,7 +53,7 @@ use Goteo\Payment\Payment;
 								</div>
 								<div class="pull-right text-right">
 									<div>
-									<?= Payment::isSubscriptionMethod($invest->method) ? $this->text('project-invest-from-subscription') : $this->text('project-invest') ?>
+									<?= $invest->invest->getMethod()->isSubscription() ? $this->text('project-invest-from-subscription') : $this->text('project-invest') ?>
 									</div>
 									<div class="amount">
 									<?= amount_format($invest->amount) ?>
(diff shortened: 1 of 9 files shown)


Resources/templates/responsive/project/participate.php
```diff
@@ -1,4 +1,8 @@
-<?php $this->layout('project/layout') ?>
+<?php
+
+use Goteo\Payment\Payment;
+
+ $this->layout('project/layout') ?>
```
davidbeig: I think i would prefer to have a function inside `Invest.php` that can tell us if it's a subscription or not. This way we could not import any class and just use `$invest->isSubscriptionMethod()`

```php
public function isSubscriptionMethod() 
{
  return self::METHOD_STRIPE_SUBSCRIPTION == $this->method
}
```

How do you see it?

davidbeig (changes requested):

Resources/templates/responsive/project/participate.php
subiabre: `Invest`s are a byproduct of a `Payment`, which can be of many shapes and types, the logic for checking if a payment method is subscription-based or not is, by design, more complex than just checking for Stripe as we might have more subscription-based payment options in the future.

I chose to leave that logic outside of the `Invest` object and into the `Payment` class to help with a separation of concerns and not enlarge an entity that is already thousand of code lines long. Hope you can agree this solution is "the less bad one".

davidbeig: I don't know... `Invest` already has functions to ask about the payments methods involved. I know the solution I posted is not scalable.

Since we can ask the Invest for their payment method, we could then ask it if it's a subscription method.

```php
<?= $invest->getMethod()->isSubscriptionMethod() ? $this->text('project-invest-from-subscription') : $this->text('project-invest') ?>
```

```php AbstractPaymentMethod
   public function isSubscriptionMethod(): bool
   {
       return Payment::isSubscriptionMethod($this->getIdNonStatic());
   }
```

I just don't really like seeing imports inside the views tbh.

Still, if you don't see it, we can stick with the development you did.

translations/ca/project.yml
```diff
@@ -23,6 +23,7 @@ project-help-license: 'Necessito ajuda per definir els retorns col·lectius i le
 project-hide-donors: 'Amagar cofinançadors'
 project-hide-needs: 'Amagar llistat de necessitats'
 project-invest: Aportació
+project-invest-from-subscription: Suscripció
```
davidbeig: This one is [subscripció](https://dlc.iec.cat/Results?IdE=0081896&DecEntradaText=subscripci%C3%)

davidbeig (changes requested):

Resources/templates/responsive/project/participate.php
subiabre: I like this one too. Will implement.

davidbeig (approved):
Well done!

PR #596, opened 2024-01-18
Fix/stripe refunds

subiabre:
#### :tophat: What? Why?
This PR fixes an issue on the Stripe Gateway that was making it impossible for the invests to be written with the correct reference value, the subscription id, under the `payment` property, making it impossible for the webhook controller to find the referenced invests when receiving a refund event.

It also saves the Stripe invoice id under the `transaction` property, this makes invests be tied 1:1 to an specific stripe invoice (i.e a payment for the subscription).

#### :pushpin: Related Issues
- Fixes https://app.asana.com/0/1206149597772083/1206194873937420

#### Testing
1. Set the stripe-cli to forward webhook events to your localhost.
2. Generate a subscription from Goteo.
3. Check on the /admin board that the new invest has the subscription id and invoice id under the payment and transaction.
4. Make a refund of the payment at the stripe dashboard.
5. Check on the /admin board that the new invest now has an status of "cancelled".

:hearts: Thank you!

### DIFF
--- a/src/Goteo/Controller/StripeSubscriptionController.php
+++ b/src/Goteo/Controller/StripeSubscriptionController.php
@@ -18,8 +18,9 @@ use Goteo\Model\Invest;
 use Goteo\Model\User;
 use Goteo\Payment\Method\StripeSubscriptionPaymentMethod;
 use Goteo\Repository\InvestRepository;
+use Stripe\Charge as StripeCharge;
 use Stripe\Event;
-use Stripe\Invoice;
+use Stripe\Invoice as StripeInvoice;
 use Stripe\StripeClient;
 use Stripe\Webhook;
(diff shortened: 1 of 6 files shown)


davidbeig (approved):
Thank you very much @subiabre !

This PR fixes what it's supossed to. I just have some questions, but the PR will de approved.

Since the StripeSubscription is not refundable, we can't return any donation via our admin pannel.

```php
    public function refundable(): bool
    {
        return false;
    }
```
This forces that the returns have to be made from the Stripe dashboard. What will happen if a project fails, or for some reason we have to massively return donations? Also, the state that the invest changes to is STATUS_CANCELLED (2) and not STATUS_RETURNED (4).

I guess that we can keep status 4 for the cases when a project fails and is not the user that triggers the return of the invest.

Thanks again for the good work!

PR #597, opened 2024-01-19
Opt-out of platform tip

subiabre:
#### :tophat: What? Why?
Sets the "platform tip" checkbox as checked by default.
See https://app.asana.com/0/1200623515637644/1206382893333183/f

:hearts: Thank you!

### DIFF
--- a/Resources/templates/responsive/invest/partials/reward_box.php
+++ b/Resources/templates/responsive/invest/partials/reward_box.php
@@ -71,7 +71,7 @@ $available = $reward->available();
                             <input type="number"
                                    class="form-control input-amount"
                                    name="donate_amount"
-                                   value="2"
+                                   value="<?= $this->donate_amount ? $this->donate_amount : amount_format($this->get_config('donate.tip'), 0, true) ?>"
                                    id="donate_amount-<?= $reward->id ?>"
                                    style="display:none;">
                         </div>
--- a/Resources/templates/responsive/invest/partials/reward_box.php
+++ b/Resources/templates/responsive/invest/partials/reward_box.php
@@ -68,6 +68,12 @@ $available = $reward->available();
                                    title="<?= $this->t('regular-amount') ?>"
                                    aria-labelledby="amount-<?= $reward->id ?>"
                                    required>
+                            <input type="number"
+                                   class="form-control input-amount"
+                                   name="donate_amount"
+                                   value="2"
+                                   id="donate_amount-<?= $reward->id ?>"
+                                   style="display:none;">
                         </div>
                         <div class="col-sm-5 col-sm-offset-1 reward-button">
                             <button type="submit" class="btn btn-block btn-success col-xs-3 margin-2"><?= $this->text('invest-button') ?></button>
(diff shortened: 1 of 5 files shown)


Resources/templates/responsive/invest/partials/reward_box.php
```diff
@@ -68,6 +68,12 @@ class="form-control input-amount"
                                    title="<?= $this->t('regular-amount') ?>"
                                    aria-labelledby="amount-<?= $reward->id ?>"
                                    required>
+                            <input type="number"
+                                   class="form-control input-amount"
+                                   name="donate_amount"
+                                   value="2"
```
davidbeig: This value is present in the configuration yaml of the application. This should be changed to search it in the configuration file through the $this->get_config() function.

src/Goteo/Controller/InvestController.php
```diff
@@ -91,8 +91,14 @@ private function validate(
 
         $custom_amount = Currency::amountInverse($amount_original, $currency);
         $project_categories = Project\Category::getNames($project_id);
+        $donate_amount = Currency::amount(2, Currency::get(Config::get('currency')));
```
davidbeig: Same about the Config::get('donate.tip')

davidbeig (changes requested):

src/Goteo/Controller/InvestController.php
```diff
@@ -91,8 +91,17 @@ private function validate(
 
         $custom_amount = Currency::amountInverse($amount_original, $currency);
         $project_categories = Project\Category::getNames($project_id);
+        $donate_amount = Currency::amount(
+            Config::get('donate.tip_amount'),
+            Currency::get(Config::get('currency'))
```
davidbeig: The execution of these lines throws an error since the Currency::get() function returns an array, that the Currency::amount() is not expecting.

davidbeig (changes requested):
https://github.com/GoteoFoundation/goteo/blob/eb2016a3da5074b5e6903984d333cf3830158834/src/Goteo/Controller/InvestController.php#L308

Here we should force to have the default value for the tip if we want it to be opt-out, if not, when a user reaches this step in the donation process it will not show by default.

src/Goteo/Controller/InvestController.php
```diff
@@ -302,22 +302,22 @@ public function signupAction($project_id, Request $request)
...
     {
         $amount = $request->query->get('amount');
-        $donate_amount = $request->query->get('donate_amount');
+        $donate_amount = $request->query->getInt('donate_amount', Config::get('donate.tip_amount'));
         $email = $request->query->has('email');
         $reward = $this->validate($project_id, $request->query->get('reward'), $amount, null, 'auto');
 
         if(!($this->skip_login && $email) && !Session::isLogged()) {
-            return $this->redirect('/invest/' . $this->project->id . '/signup?' . $this->query);
+            return $this->redirect('/invest/' . $project->id . '/signup?' . $this->query);
```
davidbeig: $project does not exist here yet, we should use $project_id and then we are done.

davidbeig (changes requested):

davidbeig (approved):
Perfect!

PR #609, opened 2024-06-20
[feat] Antispam forms

subiabre:
#### :tophat: What? Why?
Adds trap fields in forms that are invisible to humans and as such will be left blank by them, but that robots might fill.

#### :pushpin: Related Issues
https://app.asana.com/0/1205580187188381/1207505525955578

#### Testing
1. Update your DB: `migrate up`
2. Go to the `/contact` form.
3. The field is hidden below the regular email input. You can reach it by hitting the `Tab` key once when focused on the email input field. Introduce any value.
4. Application will present you with a regular success message. But if you check MailHog nothing should show up. `select * from form_honeypot;` should show a new record with the value you previously introduced.

:hearts: Thank you!

### DIFF
--- a/db/migrations/20240620092323_goteo_form_honeypot.php
+++ b/db/migrations/20240620092323_goteo_form_honeypot.php
@@ -37,7 +37,8 @@ class GoteoFormHoneypot
                 `id` BIGINT(20) UNSIGNED NOT NULL AUTO_INCREMENT,
                 `trap` VARCHAR(50) NOT NULL,
                 `prey` TEXT,
-                `template` TEXT,    
+                `template` TEXT,
+                `datetime` TIMESTAMP,
                 PRIMARY KEY (`id`)
             ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
             ";
--- a/db/migrations/20240620092323_goteo_form_honeypot.php
+++ b/db/migrations/20240620092323_goteo_form_honeypot.php
@@ -1,0 +1,55 @@
+<?php
+
+/**
+ * Migration Task class.
+ */
+class GoteoFormHoneypot
+{
+    public function preUp()
+    {
+        // add the pre-migration code here
+    }
+
+    public function postUp()
+    {
+        // add the post-migration code here
+    }
+
+    public function preDown()
+    {
+        // add the pre-migration code here
+    }
+
+    public function postDown()
+    {
+        // add the post-migration code here
+    }
+
+    /**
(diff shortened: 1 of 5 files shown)


src/Goteo/Model/FormHoneypot.php
```diff
@@ -0,0 +1,75 @@
...
+        ];
+
+        return $honeypot;
+    }
+
+    /**
+     * Checks if something got caught in the trap
+     * @return bool `true` if caught something, `false` if not
+     */
+    public static function checkTrap(string $trap, $data): bool
```
davidbeig, line 73: Why not type $data as Request?

src/Goteo/Model/FormHoneypot.php
```diff
@@ -0,0 +1,75 @@
...
+
+    public $params;
+
+    /**
+     * The template for the trap field
+     */
+    public $template = 'partials/form/honeypot';
+
+    protected $Table = 'form_honeypot';
+    static protected $Table_static = 'impact_item';
```
davidbeig:
```suggestion
    static protected $Table_static = 'form_honeypot';
```

db/migrations/20240620092323_goteo_form_honeypot.php
```diff
@@ -0,0 +1,55 @@
...
+     * @return string The SQL string to execute for the Up migration.
+     */
+    public function getUpSQL()
+    {
+        return "
+            CREATE TABLE `form_honeypot` (
+                `id` BIGINT(20) UNSIGNED NOT NULL AUTO_INCREMENT,
+                `trap` VARCHAR(50) NOT NULL,
+                `prey` TEXT,
+                `template` TEXT,
```
davidbeig: Should we create a datetime field? So that we can get info about when we have a spike on spam?

davidbeig (changes requested):
Works great!

I just have some comments, you'll find them in the code. Mainly I think that it would be interesting to know when a bot felt in the trap, what do you think about it?

After changes we can approve it :)

src/Goteo/Model/FormHoneypot.php
subiabre: I'm unsure about the scalability of strongly typing the request data here. I thought there could be other instances of forms that do not use an instance of Symfony's Request. While dirty, this gives us a faster way to add traps in any place without having to incur into refactorings of unknown scale.

davidbeig (approved):
Perfect!

Thanks <3

PR #611, opened 2024-06-27
[FIX] Sizing of honeypot input fields not true 0

subiabre:
#### :tophat: What? Why?
After careful observation it was discovered the honeypot input field was not of size 0, but it actually had a 1px border that made it take a total of 2px in the form. Making the true email input above it very slightly out of grid. This PR fixes the error and unifies the styling for the rest of the invisible honeypot HTML elements.

:hearts: Thank you!


davidbeig (approved):
Sorry for the delay!

Approved.

PR #613, opened 2024-07-08
Fix transactions id in payments to pool

davidbeig:
#### :tophat: What? Why?
*Please describe your pull request.*
Fix a bug in the donation process to the wallet of the user. Since the upgrade to use the project id in the metadata of the transactions inside the different payment providers, we are facing this bug that does not allow users to donate to their own wallet.

:hearts: Thank you!

### DIFF
--- a/src/Goteo/Payment/Method/PaypalPaymentMethod.php
+++ b/src/Goteo/Payment/Method/PaypalPaymentMethod.php
@@ -28,9 +28,11 @@ class PaypalPaymentMethod extends AbstractPaymentMethod
         /** @var ExpressGateway */
         $gateway = $this->getGateway();
         $invest = $this->getInvest();
-        $project = Project::get($invest->project);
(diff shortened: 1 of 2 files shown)


src/Goteo/Payment/Method/PaypalPaymentMethod.php
```diff
@@ -28,8 +28,12 @@ public function purchase(): ResponseInterface
         /** @var ExpressGateway */
         $gateway = $this->getGateway();
         $invest = $this->getInvest();
-        $project = Project::get($invest->project);
-        $transactionId = sprintf("%s-%s", $project->getNumericId(), $invest->id);
```
subiabre, line 32: Fixed this to pad the transaction id with leading "0". This is to avoid confusion with past invests (pre project numeric id) that did go to a project but did not include the project numeric id in the reference number, this way we can still distinguish non-project invests while keeping records sorted.

subiabre (approved):

PR #614, opened 2024-07-10
[FIX] Product references on Stripe for wallet payments

subiabre:
#### :tophat: What? Why?
A bug was introduced on https://github.com/GoteoFoundation/goteo/pull/613 where it would incorrectly try to read references to undefined `$project` for payments to the user wallet via Stripe.


davidbeig (approved):
Checked and it works as expected! 

Thank you!

PR #617, opened 2024-09-02
[fix] $dataSets variable naming

subiabre:
#### :tophat: What? Why?
Template at `Resources/templates/responsive/partials/components/data_sets.php` was expecting to be injected a `dataSets` prop to walk via foreach. However in `Resources/templates/responsive/channel/layout.php` this template was being inserted with no `dataSets` prop, instead it was being inserted with a `dataSetsSection` prop, which caused the foreach loop to fail with an `Invalid argument` error.

#### Testing
This error was being triggered mainly at /channel/feminismos/social-account-telegram. Call to this address, check that the Invalid argument error appears in your logs, then call this address with the updated code and the error will not appear anymore.

:hearts: Thank you!

### DIFF
--- a/Resources/templates/responsive/channel/layout.php
+++ b/Resources/templates/responsive/channel/layout.php
@@ -77,8 +77,8 @@ $background = $this->channel->owner_background;
 
 <?php endif; ?>
 
-<?php if ($dataSetsSection = $this->nodeSections['data_sets']): ?>
-    <?= $this->insert("channel/partials/data_sets", ['dataSetsSection' => $dataSetsSection]); ?>
+<?php if ($dataSets = $this->nodeSections['data_sets']): ?>
+    <?= $this->insert("channel/partials/data_sets", ['dataSets' => $dataSets]); ?>
 <?php endif; ?>
 
 <?= $this->insert("channel/partials/posts_section") ?>
--- a/Resources/templates/responsive/channel/layout.php
+++ b/Resources/templates/responsive/channel/layout.php
@@ -77,8 +77,8 @@ $background = $this->channel->owner_background;
 
 <?php endif; ?>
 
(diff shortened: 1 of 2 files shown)


davidbeig (changes requested):
This is not the problem with that warning.

First, let's see how the uri is `/channel/feminismos/social-account-telegram`. The `channel` refers to the `ChannelController`, then `feminismos` is the id of one of the channels. So, what is `social-account-telegram`? That i don't know, but it falls inside the route for `channel-list-projects` and the `listProjectsAction`.

That action is the one responsible to show the different projects published inside a channel and filter them by type. The `social-account-telegram` should be a type of project, but it's not. The action should redirect to the base type, that it's `available`.

Still, that has nothing to do with the error, but it's something we could also fix now.

The variable that you are changing, the `$dataSetsSection` is used to know if this Channel has a `NodeSection` for Data Sets, that's why it comes from the `$nodeSections` variable.

Inside the `channel/partials/data_sets` we can see how it's used.

```php
    $section = $this->dataSetsSection;
    if ($section):
```

And where is the `$dataSets` used? It's used in `components/data_sets`. The problem here, is that the `listProjectAction` does not give that variable to the view.

We should add it, as in the indexAction:

```php
$dataSetsRepository = new DataSetRepository();
$dataSets = $dataSetsRepository->getListByChannel([$id]);
```
You can check then that it works fine with it.

Thank you for taking the time to look into this!

subiabre:
@davidbeig reverted my previous changes and applied fixes based on your feedback. I want to thank you for preventing me from polluting the code with an apparently working, but fundamentally flawed, bugfix.

Did not fix the redirection to the `/available` base type however, as it is out of the original scope of this fix (reduce error logs) and I'm unsure about whether we should redirect or show a not found message.

davidbeig (approved):
Thanks @subiabre, it works great now :)

PR #618, opened 2024-09-02
[fix] Move ajax-loader.gif to expected route

subiabre:
#### :tophat: What? Why?
Moved the ajax-loader.gif asset to the path where it is specified to be.

:hearts: Thank you!


davidbeig (approved):
Thx! <3

PR #627, opened 2024-11-26
Add redirect capabilities to Invest Controller

davidbeig:
#### :tophat: What? Why?
*Please describe your pull request.*

Allow the process of invest to return to an outside link.

#### Testing
*Describe the best way to test or validate your PR.*

Add the domain you want to be valid to your settings.yml under the 'url' field.

```yaml
url:
    allowed_urls:
        - duckduckgo.net
```

Add a &redirect_to param to the query when accessing the invest landing of any project. Follow the donation flow and get redirected to that url. The host of that url has to match one of the domains configured in the settings, if it's not the redirect won't work.

:hearts: Thank you!

### DIFF
--- a/src/Goteo/Controller/InvestController.php
+++ b/src/Goteo/Controller/InvestController.php
@@ -101,7 +101,7 @@ class InvestController extends Controller {
             $return_to = '';
 
             if ($request->query->has('return_to')) {
-                $return_to = $request->query->get('return_to');
+                $return_to = rawurldecode($request->query->get('return_to'));
 
                 if (Domain::isAllowedDomain($return_to))
                     Session::store('return_to', $return_to);
--- a/src/Goteo/Controller/InvestController.php
+++ b/src/Goteo/Controller/InvestController.php
@@ -97,8 +97,11 @@ class InvestController extends Controller {
         );
 
         if ($request) {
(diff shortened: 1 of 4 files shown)


subiabre (dismissed):

davidbeig:
There are some things missing in this PR.

- [x] Array map for the list of allowed domains.
- [x] Move the check of allowed domains before saving it to the session of the user.
- [x] Treat the url encode in the return_to param.

davidbeig:
@subiabre whenever you can, please

src/Goteo/Controller/InvestController.php
```diff
@@ -96,11 +97,23 @@ private function validate(
             Config::get('currency')
         );
 
+        if ($request) {
+            $return_to = '';
+
+            if ($request->query->has('return_to')) {
+                $return_to = rawurldecode($request->query->get('return_to'));
```
subiabre: Why not place the url decoding inside the `isAllowedDomain` static method? This way we can safely use the functionality without having to worry about the string we are passing to it.

davidbeig: Good catch, thanks!

subiabre (approved):
LGTM!

PR #633, opened 2025-01-21
Add Mexican pesos as currency

davidbeig:
#### :tophat: What? Why?
We are going to have a mexican campaign and we need to be able to show the amount in Mexican Pesos.

#### Testing
*Describe the best way to test or validate your PR.*
Enter to any project, change the currency to Mexican Pesos and see that it's shown with the new currency, and that the amount is correct.

### :camera: Screenshots
image

:hearts: Thank you!

### DIFF
--- a/Resources/currencies.yml
+++ b/Resources/currencies.yml
@@ -20,3 +20,10 @@ GBP:
     thousands: ','
     decimal: .
     active: true
+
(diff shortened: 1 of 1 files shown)


subiabre (approved):
Works!

